PR #50, opened 2025-09-15
Pull permissions into own server

benbrandt:
This allows us to use the SDK version of an MCP server for all other
tool calls. Ideally the permission server goes away eventually.

Signed-off-by: Ben Brandt <benjamin.j.brandt@gmail.com>


benbrandt:
Not ready for merge because for some reason on the second message you send, Claude Code initializes the SDK server again and has duplicate tools.... Seems like a bug on their end

PR #51, opened 2025-09-15
Support /compact

benbrandt:
It would be nice to also support sending status via hooks when the compact action is happening (and maybe show something in the UI to show where compacting happened in the thread) but for now this at least allows it.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -298,7 +298,7 @@ export class ClaudeAcpAgent implements Agent {
             typeof message.message.content === "string" &&
             message.message.content.includes("<local-command-stderr>")
           ) {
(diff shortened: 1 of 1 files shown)


src/acp-agent.ts
```diff
@@ -284,6 +284,24 @@ export class ClaudeAcpAgent implements Agent {
...
+          ) {
+            console.log(message.message.content);
+            break;
+          }
+
+          if (
+            typeof message.message.content === "string" &&
+            message.message.content.includes("<local-command-stderr>")
+          ) {
+            console.log(message.message.contents);
```
benbrandt, line 301: Should we be returning this error to the user?

PR #85, opened 2025-10-07
Improve setSessionMode error handling for SDK messages

ddoemonn:
When `default`, `acceptEdits`, or `plan` modes fail, users see the real error from the SDK instead of a generic message.

`Bypass permissions` gets special treatment - it shows both the SDK error AND explains it needs proper configuration. I couldn't test SDK thats why I added special treatment for bypass permissions. If you can test it we can talk further.

Related to: https://github.com/zed-industries/zed/pull/39524 @benbrandt

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -383,10 +383,22 @@ export class ClaudeAcpAgent implements Agent {
       case "bypassPermissions":
       case "plan":
         this.sessions[params.sessionId].permissionMode = params.modeId;
-        await this.sessions[params.sessionId].query.setPermissionMode(params.modeId);
+        try {
+          await this.sessions[params.sessionId].query.setPermissionMode(params.modeId);
+        } catch (error) {
+          const errorMessage =
+            error instanceof Error && error.message ? error.message : "Invalid Mode";
+          if (params.modeId === "bypassPermissions") {
+            throw new Error(
+              `Failed to enable bypass permissions mode. This mode requires proper Claude Code configuration. Original error: ${errorMessage}`,
+            );
+          }
+
+          throw new Error(errorMessage);
+        }
         return {};
       default:
-        throw new Error("Invalid mode");
+        throw new Error("Invalid Mode");
     }
   }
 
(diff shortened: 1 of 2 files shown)


src/tests/acp-agent.test.ts
```diff
@@ -769,6 +769,22 @@ describe("tool conversions", () => {
   });
 });
 
+describe("setSessionMode bypass permissions error handling", () => {
```
benbrandt, line 772: I'm not sure this test adds much. I can see how it was helpful to learn the behavior, but I feel we should pull this into a helper function, or just remove the test (totally fine with the latter by the way: it helped you learn, but we don't need to spend CI cycles testing it forever)

src/acp-agent.ts
```diff
@@ -383,10 +383,22 @@ export class ClaudeAcpAgent implements Agent {
...
         this.sessions[params.sessionId].permissionMode = params.modeId;
-        await this.sessions[params.sessionId].query.setPermissionMode(params.modeId);
+        try {
+          await this.sessions[params.sessionId].query.setPermissionMode(params.modeId);
+        } catch (error) {
+          const errorMessage =
+            error instanceof Error && error.message ? error.message : "Invalid Mode";
+          if (params.modeId === "bypassPermissions") {
+            throw new Error(
+              `Failed to enable bypass permissions mode. This mode requires proper Claude Code configuration. Original error: ${errorMessage}`,
```
benbrandt, line 393: Can we give the user a hint on what they need to do here? Or does the original error message tell them?

benbrandt (changes requested):
A great start, thanks for tackling this!

src/acp-agent.ts
ddoemonn: Actually, I am trying to understand the SDK behavior - I need to do a little bit of work here. I will make the necessary changes!

ddoemonn:
@benbrandt I can finally catch the original error message from the SDK:

```
{
  code: -32603,
  message: 'Internal error',
  data: {
    details: 'Failed to enable bypass permissions mode. This mode requires proper Claude Code configuration. Original error: Cannot set permission mode to bypassPermissions since it is not available'
  }
}
```

As you can see, the original error from the SDK is:
```
Cannot set permission mode to bypassPermissions since it is not available
```

Now I need guidance from you about the error message - should I also add a default message or directly show the original error message from the SDK?

PR #86, opened 2025-10-07
Enable setting CLAUDE_CODE_EXECUTABLE to override the claude code CLI used

### LINKED ISSUE #84: Enable overriding path to the claude executable
## Background 
This ACP adapter currently always uses the claude code bundled with the SDK (@anthropic-ai/claude-agent-sdk). There's no way to specify a custom Claude Code executable, which limits flexibility for users who want to use custom wrappers over the claude code SDK. This is a fairly common pattern at peer companies in the industry.

## Proposal
Enable setting a custom claude executable using an environment variable. For example `CLAUDE_CODE_EXECUTABLE`. If not specified, the current behavior applies.

The Claude Agent SDK already supports overriding the executable this via the `pathToClaudeCodeExecutable` option in its Options interface. It just needs to be wired up.

I'm willing to throw up a PR :)

iamthebot:
## Description
Closes #84 by allowing users to set `CLAUDE_CODE_EXECUTABLE` to override the binary that's used by the claude code SDK. This is key in enterprise environments where claude CLI has been shimmed in some way.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -179,6 +179,9 @@ export class ClaudeAcpAgent implements Agent {
       // note: although not documented by the types, passing an absolute path
       // here works to find zed's managed node version.
       executable: process.execPath as any,
+      ...(process.env.CLAUDE_CODE_EXECUTABLE && {
+        pathToClaudeCodeExecutable: process.env.CLAUDE_CODE_EXECUTABLE,
+      }),
     };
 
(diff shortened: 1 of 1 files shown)


benbrandt (approved):
@iamthebot thanks for taking the time to do this! Will get this out in the next release

benbrandt:
This will be released in 0.5.4 thanks again!

PR #91, opened 2025-10-11
Add support for appending or overriding the system prompt

### LINKED ISSUE #90: Support overriding the system prompt
I'm working on an integration for [obsidian](https://github.com/WhiskeyJack96/obsidian-acp) and it works well but because this isnt a traditional "code editor" the system prompt isn't ideal. When running in the terminal I can append to the system prompt or use a subagent when needed but Ideally I'd like to be able to adjust the system prompt at startup when running in the obsidian context. 

I see that currently the system prompt is set to the claude code system prompt, I would like to propose that this integration support a `_meta` field of `system_prompt` which would allow overriding the system prompt of the type `string | { type: 'preset'; preset: 'claude_code'; append?: string }`. This would enable none traditional integrations to specify the rules of engagement and would also enable editors to append editor specific idiosyncrasies to the agent.

Happy to do this work if theres interest/folks are willing to see it done!

WhiskeyJack96:
resolves #90 

Tested both the `append` and `override` flows by sending:
```javascript
const resp = await this.connection.newSession({
			cwd: basePath,
			_meta: {
				systemPrompt: "you are an obsidian assistant agent, you should be helpful to the user when working with their notes"
			},
			mcpServers: []
		});
```

```javascript
const resp = await this.connection.newSession({
			cwd: basePath,
			_meta: {
				systemPrompt: {
					append: "always speak in spanish"
				}
			},
			mcpServers: []
		});
```

and confirmed that in both cases those prompts are what gets sent to the agent!

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -169,19 +169,24 @@ export class ClaudeAcpAgent implements Agent {
       },
     };
 
-    let systemPrompt: Options["systemPrompt"] =  { type: "preset", preset: "claude_code" }
-    if (params._meta?.systemPrompt){
-      if (typeof params._meta.systemPrompt === "string") {
-        systemPrompt = params._meta.systemPrompt
-      } else {
-        systemPrompt = {...systemPrompt, ...params._meta.systemPrompt}
+    let systemPrompt: Options["systemPrompt"] = { type: "preset", preset: "claude_code" };
+    if (params._meta?.systemPrompt) {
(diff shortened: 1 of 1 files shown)


WhiskeyJack96:
@cla-bot check

benbrandt:
@WhiskeyJack96 this looks great! I made the checks slightly stricter to be safe, but otherwise it makes sense to me!

PR #92, opened 2025-10-12
Make write and edit permissions consistent

WhiskeyJack96:
Currently edit requires a permissions call and write does not. This means that if clients want to allow users to approve `Write` calls they need to add handling to `HandleFileWrite` but if they do that then edits get _two_ permissions checks.

Alternatively we could remove the write tool from approved tools (which would make more sense to me) but I'm assuming its there for a good reason, if not happy to change to remove it instead!

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -191,7 +191,6 @@ export class ClaudeAcpAgent implements Agent {
       disallowedTools.push("Read");
     }
     if (this.clientCapabilities?.fs?.writeTextFile) {
-      allowedTools.push(toolNames.write, toolNames.edit);
       disallowedTools.push("Write", "Edit");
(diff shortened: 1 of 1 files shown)


benbrandt:
@WhiskeyJack96 thanks for reporting. I pushed a change to not push Write to allowed, since I'm not sure that is the desired behavior anyway... 
Do you want to test how that works for your integration? And I'll do the same on my end

WhiskeyJack96:
This works as expected (my integration now requires two approvals for both edit and write)!

benbrandt:
Great. It is working as expected for me as well, and we already had handling in the permissions tool that if the user allows edits than we permit the write tool, but if not, then they have to approve.

Thanks for catching!

PR #108, opened 2025-10-17
Don't allow bypassPermissions to be enabled if in root/sudo mode, as Claude Code won't start

### LINKED ISSUE #104: Session initialization still fails in v0.6.3 (regression persists from v0.6.1)
## Summary

Sessions fail to initialize in v0.6.3. While PR #99 attempted to fix the issue introduced in v0.6.1, the problem persists.

### v0.6.0 (Working)
- Sessions initialize successfully

### v0.6.3 (Broken)
- Sessions fail to initialize
- Error: `Request session/new timed out`

### Current Code (v0.6.3)

```typescript
const q = query({
  prompt: input,
  options,
});
const models = await getAvailableModels(q);
const permissionMode = "default";
// Change it back to default
await q.setPermissionMode(permissionMode);
```

The fix added `await getAvailableModels(q)` before `setPermissionMode()` to ensure initialization completes. However, `getAvailableModels()` itself sends a control request:

```typescript
async function getAvailableModels(query: Query): Promise<SessionModelState> {
  const models = await query.supportedModels();

  const currentModel = models.find((model) => model.value === DEFAULT_MODEL_ID) || models[0];
  await query.setModel(currentModel.value);  // Sends control request immediately after supportedModels()

  // ...
}
```

## (Potential) Root Cause Analysis

The sequence in v0.6.3 becomes:

1. `query()` - spawns process, starts async initialization
2. `supportedModels()` - waits for initialization
3. `setModel()` - sends control request immediately
4. `setPermissionMode()` - sends another control request immediately

Both `setModel()` and `setPermissionMode()` use the `request()` method which immediately writes to the child process:

```typescript
request(request) {
  const requestId = Math.random().toString(36).substring(2, 15);
  // ...
  return new Promise((resolve, reject) => {
    this.pendingControlResponses.set(requestId, (response) => {
      // Wait for response
    });
    this.transport.write(JSON.stringify(sdkRequest) + '\n');
  });
}
```

**The Pattern**: While `supportedModels()` waits for initialization (`await this.initialization`), both `setModel()` and `setPermissionMode()` do not. Multiple control requests are sent in rapid succession before the child process is ready to handle them.

SDK methods that wait for initialization:
```typescript
async supportedCommands() {
  return (await this.initialization).commands;
}

async supportedModels() {
  return (await this.initialization).models;
}
```

SDK methods that don't wait:
```typescript
async setPermissionMode(mode) {
  await this.request({ subtype: "set_permission_mode", mode });
}

async setModel(model) {
  await this.request({ subtype: "set_model", model });
}
```

## Environment

- Package: `@zed-industries/claude-code-acp`
- Affected versions: v0.6.3
- Working version: v0.6.0

benbrandt:
Closes #104

Turns out this was the root cause, not the order of messages after all.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.6.6
+
+- Do not enable bypassPermissions mode if in root/sudo mode, because Claude Code will not start
+
 ## 0.6.5
 
 - Fix for duplicated text content after streaming
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@zed-industries/claude-code-acp",
-  "version": "0.6.5",
+  "version": "0.6.6",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@zed-industries/claude-code-acp",
-      "version": "0.6.5",
+      "version": "0.6.6",
       "license": "Apache-2.0",
       "dependencies": {
         "@agentclientprotocol/sdk": "0.4.8",
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
   "bin": {
     "claude-code-acp": "./dist/index.js"
   },
-  "version": "0.6.5",
+  "version": "0.6.6",
   "description": "An ACP-compatible coding agent powered by the Claude Code SDK (TypeScript)",
   "main": "dist/index.js",
   "type": "module",
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -85,6 +85,9 @@ type ToolUseCache = {
 
 const DEFAULT_MODEL_ID = "default";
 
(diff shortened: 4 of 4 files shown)


priteshgupta:
Btw there is an undocumented workaround as well by setting the "IS_SANDBOX" env variable. Adding `ENV IS_SANDBOX=true` makes the repro setup work with the newer pre fix packages versions.

Not sure if this is something the adapter should support, but it can be a potential issue for setups using that mode and relying on that env var to bypass the error.

This seems to be moderately known: https://www.google.com/search?q=%22claude+code%22+%22IS_SANDBOX%22.

Uzay-G:
Hi, I would like to use that IS_SANDBOX=1 feature with ACP. it's pretty commonly used in headless modes, and quite valuable to support.

Uzay-G:
I am trying to use ACP via claude on a remote machine

PR #121, opened 2025-10-24
prevent invalid locations when read call has no path

SteffenDE:
The ACP spec has the path as a required field for tool call locations, but I've seen Claude generate read calls without an input path, which would generate:

```json
{
  "content": [],
  "kind": "read",
  "locations": [
    {
      "line": 0
    }
  ],
  "rawInput": {},
  "sessionUpdate": "tool_call",
  "status": "pending",
  "title": "Read File",
  "toolCallId": "toolu_012PzeR5whUjWN4ij2HBubJw"
}
```

And then fail to parse as SessionNotification in the SDK.

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -113,12 +113,14 @@ export function toolInfoFromToolUse(
       return {
         title: "Read " + (input.file_path ?? "File") + limit,
         kind: "read",
-        locations: [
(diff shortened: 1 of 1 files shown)


SteffenDE:
@cla-bot check

PR #124, opened 2025-10-25
Allow for the package to be used as a dependency, while preserving existing behavior when run

gsabran:
This change will allow for this package to be used directly, without launching a subprocess and communicating over stdio.


gsabran:
@cla-bot check

benbrandt:
@gsabran there were some cases (Zed included) that were relying on this old entrypoint. https://github.com/zed-industries/claude-code-acp/pull/127


I reverted and move the library entrypoint to a new file. Let me know if v0.8.0 works for you

gsabran:
I'll test it, but that should work. Thanks

PR #128, opened 2025-10-27
Add metadata to tool calls, including structured tool output

gsabran:
Use the `_meta` field for tool calls to forward:
- the original tool name (ie `Bash` etc)
- the structured output of the tool call

The output is received by using the `PostToolUse` hook from the SDK. The `tool_call_update` for the tool result is now only sent when both the hook callback and the original tool result (plain text) have been received.

Claude Code describes [in its documentation ](https://docs.claude.com/en/api/agent-sdk/typescript#tool-output-types)the expected structure of the output for each tools. Unfortunately those are not part of the SDK, and [empirically I've found that there was some differences](https://github.com/getcmd-dev/cmd/blob/58c3100a36f7e8ec92e1fd04b3af3091867d6cfc/local-server/src/server/endpoints/sendMessage/acp/clients/claudeCode/types.ts#L265). Still I've found that having this structured output was useful for my client to provide a richer UX. Maybe Zed will too.

I was not sure how to add tests for this change, given the current setup

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -77,15 +77,6 @@ type BackgroundTerminal =
       pendingOutput: TerminalOutputResponse;
     };
 
-export type CachedToolUse = {
-  type: "tool_use" | "server_tool_use" | "mcp_tool_use";
-  id: string;
-  name: string;
-  input: any;
-  toolResponse?: unknown;
-  output?: any;
-};
-
 /**
  * Extra metadata that the agent provides for each tool_call / tool_update update, under the `_meta.claudeCode` key.
  */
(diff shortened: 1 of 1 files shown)


benbrandt (changes requested):
We shoudn't have to cache the results, we should be able to send this as multiple tool_call_updates to the same tool call, only updating the fields we are about

gsabran:
makes sense. I'll update

gsabran:
Updated, let me know if this is what you expected. There's now a new update that is only relevant as it contains structured output data in the meta parameter.

src/acp-agent.ts
```diff
@@ -75,6 +77,16 @@ type BackgroundTerminal =
       pendingOutput: TerminalOutputResponse;
     };
 
+/**
+ * Extra metadata that the agent provides for each tool_call / tool_update update, under the `_meta.claudeCode` key.
+ */
+export type ToolUpdateMeta = {
+  /* The name of the tool that was used in Claude Code. */
+  toolName: string;
```
benbrandt: Nice this is on the list of things to add to tool calls in general. Meta makes sense for now

PR #129, opened 2025-10-29
Remove filtering of custom slash commands (MCP prompts)

### LINKED ISSUE #115: Reason for filtering MCP prompts in command list
I have seen the discussion around filtering out certain commands that could cause inconsistencies, e.g. /model. But I am wondering about why you are filtering out MCP prompts in the commands list: 
`
  .filter(
    (command: AvailableCommand) =>
      !(command.name.match(/\(MCP\)/) || UNSUPPORTED_COMMANDS.includes(command.name)),
  );
`, i.e., filtering out commands that has (MCP) in them.

What is the reason you are doing this? It seems to be a bit of a blanket filtering, right? 
I am not very knowing of how it works internally, but these commands are just injecting prompts into the conversation, I am having a hard time figuring out why these would be problematic. 

It is a bit limiting not being able to use MCP prompts in ACP-compatible clients.

Thanks!

joeljohansson99:
Custom MCP slash commands were being filtered alongside built-in commands (like /config, /agents, etc.). Since they're user-defined rather than built-in, i.e. they are just a way to inject prompts, I think we dont need to filter them out, but I am not certain. You know better since you implemented the filter in the first place. If this causes issues with MCP commands, I will happy to adjust the approach. Fixes #115

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,3 @@
 node_modules
 dist
+.DS_Store
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -553,16 +553,17 @@ async function getAvailableSlashCommands(query: Query): Promise<AvailableCommand
   return commands
     .map((command) => {
       const input = command.argumentHint ? { hint: command.argumentHint } : null;
+      let name = command.name;
+      if (command.name.endsWith(" (MCP)")) {
+        name = `mcp:${name.replace(" (MCP)", "")}`;
+      }
       return {
-        name: command.name,
+        name,
         description: command.description || "",
         input,
       };
     })
-    .filter(
-      (command: AvailableCommand) =>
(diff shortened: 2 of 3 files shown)


bennetbo:
@joeljohansson99 Hey, can you sign the CLA so we can merge this 😄?
We pushed a small change that replaces the ` (MCP)` suffix with `/mcp` since we found the UX of a slash command name with a space in between quite confusing. Should be good to merge.

joeljohansson99:
Thank you! Sounds good. I will need to have our legal department look over the agreement first. Sorry about that. I will do my best to push for a quick process.

bennetbo:
Hey! In that case, I think it would be easier to merge this as a separate PR, since we'd like to get in our changes soon. 
Thank you for bringing to our attention that MCP slash-commands are supported via the SDK and for submitting an issue/your PR.

PR #131, opened 2025-10-29
Return Claude Code's session id in the initialization request

gsabran:
Relates to the first issue discussed in https://github.com/zed-industries/claude-code-acp/issues/130

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -99,6 +99,7 @@ export class ClaudeAcpAgent implements Agent {
   fileContentCache: { [key: string]: string };
   backgroundTerminals: { [key: string]: BackgroundTerminal } = {};
   clientCapabilities?: ClientCapabilities;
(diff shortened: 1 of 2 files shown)


src/acp-agent.ts
```diff
@@ -250,6 +251,29 @@ export class ClaudeAcpAgent implements Agent {
...
+            break;
+          }
+          // Wait for the first system message with init subtype that contains session_id
+          if (message.type === "system" && message.subtype === "init") {
+            resolve(message.session_id);
+            break;
+          }
+        }
+      })();
+    });
```
gsabran: I hope this is not breaking anything. I'll do more manual testing to ensure we don't end up waiting forever on Claude Code

gsabran: Well this doesn't work. Claude Code doesn't send an event until one message has been pushed to the input.

Issue #157, opened 2025-11-14
Option to disable ALL built-in tools? (please?)


fasterthanlime:
Hello Zed friends!

For _reasons_ I would like an option to disable every single built-in ACP tool available to Claude.

I provide my own, and they keep trying the built-in ones, running afoul of sandboxing limitations etc. — denying permission is not enough because they just, won't, stop, trying. Stubborn little things.

So, yeah, a `--disable-all-builtin-mcp-tools` would be amazing — hopefully it would also amend the agent's prompt so it doesn't even learn about these built-in tools in the first place.

I'm running over to `codex-acp` to open a similar issue, so, if this reads familiar, that's why.

benbrandt:
@fasterthanlime so if I understand correctly, you basically want to disallow all tools to the agent?

We supply some tools to claude code that overwrite the built-in ones. But if I understand correctly you want it to have _no_ tools at all?
And maybe a way to amend/edit the system prompt in this mode?

Are you trying to run this in Zed? Or somewhere else? (would also help figure out how best to thread this through) thanks!

fasterthanlime:
> you basically want to disallow all tools to the agent?

that's correct. EXCEPT the ones I supply through my singular MCP server. I don't want Claude's built-in tools nor the ACP built-ins provided by claude-code-acp (sorry — the permission system and the read-before-write system are all neat! I just have something different in mind!)

> And maybe a way to amend/edit the system prompt in this mode?

that would be the cherry on top. right now I'm trying to adjust expectations in the user prompt but... it's not very effective.

> [@fasterthanlime](https://github.com/fasterthanlime) so if I understand correctly, you basically want to disallow all tools to the agent?
> 
> We supply some tools to claude code that overwrite the built-in ones. But if I understand correctly you want it to have _no_ tools at all? And maybe a way to amend/edit the system prompt in this mode?
> 
> Are you trying to run this in Zed? Or somewhere else? (would also help figure out how best to thread this through) thanks!

somewhere else, not open-source yet.

part of the reason is worktree support: the claude-code-acp process runs with a cwd of `/tmp/some-empty-dir` — every MCP tool call knows from context which worktree it's supposed to be in, and only allows operations there.

(and the other reasons are many, I do cool stuff like "agent can't apply a patch if it's gonna break rust syntax", "transparently apply formatter after edit", "use syntactic merge to help resolve conflicts", and also "notify agents when they try to step on other agents' toes" — lotsa things.)

[...]

PR #158, opened 2025-11-14
Add experimental support for disabling built-in tools via _meta flag

fasterthanlime:
Add ability to disable built-in tools (Read, Write, Edit, Bash, etc.) when `_meta.disableBuiltInTools` is set to true. This allows clients to rely solely on MCP tools without the built-in tool implementations.

---

This is just my take on #157 — not necessarily the one that makes sense /cc @benbrandt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.10.5
+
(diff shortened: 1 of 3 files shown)


fasterthanlime:
@cla-bot check

PR #163, opened 2025-11-17
Fix read/edit tool error handling so upstream errors surface

### LINKED ISSUE #162: Read/Edit tools crash instead of surfacing readTextFile errors
When `agent.readTextFile` returns `{ error: ...}` (e.g., file too large, permission denied), the Read/Edit tools assume `content` exists and try to slice it.  That throws "Cannot read properties of undefined (reading 'indexOf')", so we never see the real upstream error and the Edit tool may still attempt a write.

To reproduce, start a session and ask Claude to read a large file or a file without permission.  The agent reports a TypeError rather than reporting the error message.  Alternatively, try editing a file before granting trusted access.

We can fix this by
- adding guards that extract the upstream error and surface the message,
- skipping `writeTextFile` when the initial read fails.

I've done so in my fork, which I'll send as a PR in a bit.

ultronozm:
Fixes #162:
- guard both tool callbacks so they return the host-provided error instead of crashing on missing content
- skip writeTextFile in the edit tool when the initial read fails, mirroring the read tool behavior
- add regression tests

### DIFF
--- a/src/tests/mcp-server-read.test.ts
+++ b/src/tests/mcp-server-read.test.ts
@@ -46,9 +46,7 @@ describe("MCP server read-related tools", () => {
       limit: 2000,
     });
 
-    expect(result.content[0].text).toBe(
-      "Reading file failed: No file contents for /tmp/file.txt.",
-    );
+    expect(result.content[0].text).toBe("Reading file failed: No file contents for /tmp/file.txt.");
   });
 
   it("reports missing contents inside the edit tool and skips writes", async () => {
@@ -68,9 +66,7 @@ describe("MCP server read-related tools", () => {
       replace_all: false,
     });
 
-    expect(result.content[0].text).toBe(
-      "Editing file failed: No file contents for /tmp/file.txt.",
-    );
+    expect(result.content[0].text).toBe("Editing file failed: No file contents for /tmp/file.txt.");
     expect(agent.writeTextFile).not.toHaveBeenCalled();
   });
 });


ultronozm:
@cla-bot check

ultronozm:
@cla-bot check

src/mcp-server.ts
```diff
@@ -101,15 +150,16 @@ Usage:
             };
           }
 
-          const content = await agent.readTextFile({
+          const readResponse = await agent.readTextFile({
```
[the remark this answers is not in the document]
ultronozm: If `agent.readTextFile` throws, the existing try/catch handles it and returns "Reading file failed: ..."., as before. 

If `agent.readTextFile` resolves to { error: "..." }, then `ensureReadContent` throws with that string, which the same try/catch converts to "Reading file failed: ...".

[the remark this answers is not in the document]
benbrandt: But it shouldn't... that is handled by the SDK, so what comes back only has the content. If it got an error it would already throw

src/mcp-server.ts
```diff
@@ -253,47 +257,62 @@ Usage:
...
+            };
+          }
+
+          const readResponse = await agent.readTextFile({
+            sessionId,
+            path: input.file_path,
+          });
+
+          if (typeof readResponse?.content !== "string") {
+            throw new Error(`Claude Code did not return file contents for ${input.file_path}.`);
```
benbrandt:
```suggestion
            throw new Error(`No file contents for ${input.file_path}.`);
```

src/tests/mcp-server-read.test.ts
```diff
@@ -0,0 +1,76 @@
...
+    const readTool = getTool(server, "Read");
+
+    const result = await readTool.callback({
+      file_path: "/tmp/file.txt",
+      offset: 1,
+      limit: 2000,
+    });
+
+    expect(result.content[0].text).toBe(
+      "Reading file failed: Claude Code did not return file contents for /tmp/file.txt.",
```
benbrandt:
```suggestion
      "Reading file failed: No file contents for /tmp/file.txt.",
```

src/tests/mcp-server-read.test.ts
```diff
@@ -0,0 +1,76 @@
...
+
+    const result = await editTool.callback({
+      file_path: "/tmp/file.txt",
+      old_string: "foo",
+      new_string: "bar",
+      replace_all: false,
+    });
+
+    expect(result.content[0].text).toBe(
+      "Editing file failed: Claude Code did not return file contents for /tmp/file.txt.",
```
benbrandt:
```suggestion
      "Editing file failed: No file contents for /tmp/file.txt.",
```

PR #177, opened 2025-11-24
Logging configuration

gsabran:
Allow to provide a logger to replace `console` (`console` remains the default value)


benbrandt (approved):
Looks great, thanks!

PR #180, opened 2025-11-27
fix: Access to undefined meta field

benbrandt:


benbrandt:
cc @gsabran

gsabran:
ouch sorry.

PR #199, opened 2025-12-10
Allow to resume previous session

gsabran:
When the ACP client persists the session id to support resuming, it has no way to pass it back to the agent to resume an existing session as the current implementation always creates a new id that is now shared with Claude Code.

My client used to pass the value in the `--resume` within Claude Code option, but `--resume` is not compatible with the new `--session-id`:

```
--session-id cannot be used with --continue or --resume.
```

Change:

When `--resume` is used, extract its value and use it as the session id.


benbrandt (approved):
Thanks!

PR #205, opened 2025-12-12
Fix Node.js v18 compatibility with JSON imports

### LINKED ISSUE #204: Node.js v18 compatibility broken with JSON import assertions
## Problem

The application fails to run in Zed editor (which uses Node.js v18.19.0) with a syntax error when trying to load package.json using the JSON import assertion syntax.

## Error

```
SyntaxError: Unexpected token 'with'
```

This appears in Zed logs when attempting to initialize the ACP agent.

## Root Cause

After updating @types/node from 24.10.1 to 25.0.0 (commit 634f1ef), the code started using the JSON import assertion syntax:

```typescript
import packageJson from "../package.json" with { type: "json" }
```

The `with` keyword for import assertions is only supported in Node.js 20.10.0 and later. Zed editor bundles Node.js v18.19.0, which doesn't support this syntax.

## Details

- Affected file: `/src/acp-agent.ts`
- @types/node version that introduced the issue: 25.0.0
- Zed's Node.js version: v18.19.0
- Minimum Node.js version supporting import assertions: v20.10.0

## Impact

- ACP agent cannot initialize in Zed editor
- Users running Zed cannot use Claude Code ACP
- The application is completely broken for Zed users

## Suggested Fix

Use `createRequire` from the `node:module` package instead, which is compatible with Node.js v18:

```typescript
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);
const packageJson = require("../package.json");
```

This approach maintains ES module compatibility while working with Node.js v18.19.0.

xqsit94:
## Summary

Fixes #204

Replace JSON import assertions with `createRequire` to support Node.js v18.19.0 runtime used by Zed editor.

## Problem

The code was using `import packageJson from "../package.json" with { type: "json" }` syntax, which requires Node.js 20.10.0+. However, Zed bundles Node.js v18.19.0, causing a runtime error:

```
SyntaxError: Unexpected token 'with'
```

This issue was introduced when `@types/node` was updated to 25.0.0, which enforces the newer import attribute syntax that isn't supported by Node.js v18.

## Solution

Use `createRequire` from `node:module` to load the JSON file:

```typescript
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);
const packageJson = require("../package.json");
```

This approach:
- ✅ Works with Node.js v18.19.0
- ✅ Satisfies TypeScript with @types/node 25.0.0
- ✅ Maintains ES module compatibility

## Test Plan

- [x] TypeScript compilation passes (`tsc --noEmit`)
- [x] Build succeeds (`npm run build`)
- [ ] Tested in Zed editor with Node.js v18.19.0

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -54,8 +54,11 @@ import {
 } from "./tools.js";
 import { ContentBlockParam } from "@anthropic-ai/sdk/resources";
 import { BetaContentBlock, BetaRawContentBlockDelta } from "@anthropic-ai/sdk/resources/beta.mjs";
-import packageJson from "../package.json" with { type: "json" };
+import { createRequire } from "node:module";
 import { randomUUID } from "node:crypto";
+
+const require = createRequire(import.meta.url);
+const packageJson = require("../package.json");
 
 export const CLAUDE_CONFIG_DIR = process.env.CLAUDE ?? path.join(os.homedir(), ".claude");
 


benbrandt:
As I mentioned in this issue, this isn't an issue with the bundled Node version, and Node 18 is no longer a supported LTS version (v20 is, which supports this syntax)
going to close but thanks for looking into it

PR #207, opened 2025-12-12
feat(unstable): add session/resume support

SteffenDE:
Verified in Tidewave!

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -234,15 +234,17 @@ export class ClaudeAcpAgent implements Agent {
   }
 
   async resumeSession(params: ResumeSessionRequest): Promise<ResumeSessionResponse> {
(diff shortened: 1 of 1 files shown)


Tihomir971:
@SteffenDE How do you activate resume?
The /resume command is not supported by Claude Code.
Available commands: /compact, /init, /pr-comments, /review, /security-review, /mcp:svelte:svelte-task, /login, /logout

SteffenDE:
Resume is a protocol feature that both the agent and the client need to support. As far as I know, on the agent side only claude-code-acp supports it right now and on the client side I only know about [Tidewave](https://tidewave.ai) implementing it. I assume you’re trying to use it through Zed or some other client?

Tihomir971:
@SteffenDE yes, i was trying to use it through Zed. thanks for clarification

Issue #216, opened 2025-12-17
Can't run execuable after NPM install


doppleware:
Tried on  MacOS / Ubuntu

```
venv ❯ npm install @zed-industries/claude-code-acp


up to date, audited 100 packages in 619ms

30 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities

~/projects/z
venv ❯ claude-code-acp
zsh: command not found: claude-code-acp
```

doppleware:
Solved... For anyone encountering the same problem, needed to install into NPM globally:
```
npm install -g @zed-industries/claude-code-acp
```

PR #224, opened 2025-12-25
Change npm install command to global installation in README

ForteScarlet:
As mentioned in #216, npm installations should be global, but the documentation does not reflect this


ForteScarlet:
@cla-bot check

PR #226, opened 2025-12-30
Add MAX_THINKING_TOKENS env support and fix file-read-info format

omgpointless:
PR for awareness and discussion.

1) claude-code-acp doesn't currently support configuring "maxThinkingTokens" (https://platform.claude.com/docs/en/agent-sdk/typescript) nor is it configured in the query. Result, at least for me, is that agent thinking chunks are never flowing through the chain. I added environment support to my ACP host, added env variable to query init for maxThinkingTokens following the pyton sdk env "MAX_THINKING_TOKENS". Result is now agent thinking chunks DO flow. And I can see Claude reason. Before, no chunks ever. This is perhaps based on configuration that Claude Agent SDK reads but could be made more flexibel. Such as this suggestions.

2) Fix file-read-info message: was showing nonsensical range format 'Read lines 160-40', now shows 'Read 40 lines starting at line 160'. The way this was written was confusing for myself and for Claude. If the file-read-info is supposed to give information, then lines like "Read lines 160-40" says absolutely nothing.

Reproduce:

Configure your ACP host to parse thinking tokens with no other external configuration. Claude will not send any.
Enable maxThinkingTokens in ACP and set it to something like 8096. Try your ACP implementation again, Claude now sends thinking chunks.

Bugs:

Anthropic has a meta issue. Don't talk about the "thinking" tag with Claude. It will break output regardless of what client you use. (even on Claude Code CLI)

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -175,7 +175,7 @@ Usage:
 
             if (result.wasLimited) {
               readInfo += `Read ${result.linesRead} lines (hit 50KB limit). `;
-            } else {
+            } else if (input.offset && input.offset > 1) {
               readInfo += `Read lines ${input.offset}-${input.offset + result.linesRead}.`;
             }
 


omgpointless:
@cla-bot check

benbrandt:
@omgpointless which model are you using? As the claude agent sdk just supplies an override in this case, I would expect it to be identical to the default in other situations. If you could supply some more info that would be helpful.

Also, we should be able to supply this to the options object, and we don't have to call a method on the query after we start it

omgpointless:
@benbrandt hey. Default model in available model reports as Claude Opus (default). I.e standard Claude Code model these days.
I have not done manual configuration. My Claude Code CLI usually has "Thinking" configured on (and still does). Have been pivoting to claude-code-acp lately in my own host.

So I went on this little exploration to basically "let us confirm I _CAN_ get thinking tokens" (I am fond of it personally). And it was not until I added this configuration that "Aha. There they are!"

But yes, I did not 100% get fully into the codebase, perhaps it can be on Options instead of Query? Would love it if it was a thing or if we can better understand the behavior.

Thanks for replying :) and happy new year

PR #232, opened 2026-01-03
feat: add hook config support

### LINKED ISSUE #144: Support for hooks
I've tried (through Emacs agent-shell) using hooks and they don't seem to work at all with ACP. I'm wondering if it's a problem with the client (agent-shell), this project or the protocol/SDK?

I'm testing with:

.claude/settings.json
```json
{
  "hooks": {
    "PostToolUse": [
      {
        "matcher": "Edit|Write",
        "hooks": [
          {
            "type": "command",
            "command": "cat >> test.txt"
          }
        ]
      }
    ]
  }
}
```

Using `claude-code` directly, the `test.txt` gets populated after each edit while nothing happens with `claude-code-acp`.

PeterCardenas:
Closes https://github.com/zed-industries/claude-code-acp/issues/144

Add support for specifying `hookConfigs` when creating a session to react to hook events.

Hooks are specified with the following type:

```ts
export type HookConfig = {
  /** The hook event to listen for */
  event: HookEvent;
  /** 
   * Optional matcher string to filter which tools trigger this hook.
   * Only relevant for tool-related hooks: PreToolUse, PostToolUse, PostToolUseFailure, PermissionRequest
   */
  matcher?: string;
  /** Command to execute */
  command: string;
  /** Arguments for the command */
  args?: string[];
  /** Environment variables to pass to the command */
  env?: Record<string, string>;
};
```

These commands are executed with environment variables populated with the tool input, so arguments or the command itself can handle the input accordingly.
To give hook output, the command must print, and only print, a JSON string to stdout which will be parsed to use as the hook result. Otherwise, the agent will continue.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -47,6 +47,7 @@ import * as os from "node:os";
 import { nodeToWebReadable, nodeToWebWritable, Pushable, unreachable } from "./utils.js";
 import { createMcpServer } from "./mcp-server.js";
 import { EDIT_TOOL_NAMES, acpToolNames } from "./tools.js";
+import { HookConfig, transformHookConfigs } from "./hook-config.js";
 import {
   toolInfoFromToolUse,
   planEntries,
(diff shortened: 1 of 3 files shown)


PeterCardenas:
@cla-bot check

AlephNotation:
Checking in on this. It would be nice to support some of the popular tools that use hooks

PeterCardenas:
closing this as claude agent sdk seems to support hooks from global config now. anyone else interested feel free to reuse the patch

PR #249, opened 2026-01-17
Fix glob matching for file permissions on Windows

omgpointless:
8 tests in settings.test.ts fail on Windows because `minimatch` expects forward slashes but `path.normalize()` returns backslashes.

Converts paths to forward slashes before matching.


bennetbo:
Thank you!

PR #267, opened 2026-01-28
fix: respect user model preference from settings.json

### LINKED ISSUE #225: claude-code-acp ignores user model configuration from ~/.claude/settings.json
## Problem

`claude-code-acp` always selects the first model from `supportedModels()` array, ignoring user's model preference configured in `~/.claude/settings.json`.

## Root Cause

In `src/acp-agent.ts`, the `getAvailableModels()` function:

```typescript
async function getAvailableModels(query: Query): Promise<SessionModelState> {
  const models = await query.supportedModels();

  // Query doesn't give us access to the currently selected model, 
  // so we just choose the first model in the list.
  const currentModel = models[0];  // ❌ Always picks first model
  await query.setModel(currentModel.value);
  
  return {
    availableModels: models.map(...),
    currentModelId: currentModel.value
  };
}
```

This overrides the user's model preference that the Claude Agent SDK properly loads from settings.

## Expected Behavior

If a user has `"model": "opus"` in `~/.claude/settings.json`, claude-code-acp sessions should use Opus, not the first model in the array (typically "default").

## Suggested Fix

**Option 1: Don't override the model**
```typescript
async function getAvailableModels(query: Query): Promise<SessionModelState> {
  const models = await query.supportedModels();
  
  // Don't call setModel() - Query already has the right model from user settings
  
  return {
    availableModels: models.map(...),
    currentModelId: undefined  // Or detect which is active
  };
}
```

**Option 2: Load settings explicitly**
```typescript
// Ensure Query is created with settingSources
const q = query({
  prompt,
  options: {
    ...options,
    settingSources: ['user', 'project', 'local'],
    // Don't override model if user configured it
  }
});
```

**Option 3: Check user settings before defaulting**
```typescript
const userSettings = await loadUserSettings();
if (userSettings.model) {
  await query.setModel(userSettings.model);
} else {
  await query.setModel(models[0]);  // Only fallback if no preference
}
```

## Impact

Users who configure specific models (like `opus` or `sonnet[1m]`) in their Claude Code settings expect those preferences to be respected when using claude-code-acp through ACP clients like Zed, Neovim, or Emacs.

Related to #30 (model selection feature request).

Railly:
- Integrates model into existing settings infrastructure
- Respects precedence: user < project < local < enterprise
- Removes duplicate file reading logic

Fixes #225

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -811,7 +811,7 @@ export class ClaudeAcpAgent implements Agent {
     };
 
     const availableCommands = await getAvailableSlashCommands(q);
-    const models = await getAvailableModels(q);
+    const models = await getAvailableModels(q, settingsManager);
 
     // Needs to happen after we return the session
     setTimeout(() => {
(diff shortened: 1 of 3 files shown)


Railly:
@cla-bot check

Railly:
@cla-bot check

PR #270, opened 2026-01-28
fix: Map Claude Code tool results to valid ACP content blocks

jcarver989:
This PR fixes an issue with `claude-code-acp` where incoming tool results from the Anthropic SDK would, in several cases, get mapped to invalid ACP `ContentBlock`s inside of the `toAcpContentUpdate` method. This prevents ACP clients from being able to properly render tool calls.

Per the [ACP spec](https://agentclientprotocol.com/protocol/schema#contentblock) `ContentBlock` can only have a `type` field with a value of: `text | image | audio | resource | resource_link`. However, several of Claude Code's tools return results with different `type`'s . 

This PR adds a `toAcpContentBlock` method that maps tool call results to valid ACP `ContentBlock`. It attempts to be defensive and will map unknown types to a text content block. This avoids breakages if Anthropic changes the types upstream, or introduces new tools.


jcarver989:
@cla-bot check

jcarver989:
@cla-bot check

PR #291, opened 2026-02-05
test: add tests for cross-tool rule isolation

ElleNajt:
## Summary
- Adds tests verifying that permission rules only apply to their intended tool types
- Tests confirm the fix from #282 (commit d471ccf) works correctly

## Tests added
- Bash rules should not match Edit/Write/Read operations
- Edit rules should not match Bash operations
- Read rules should not match Bash operations

These tests catch the bug where `Bash(./**:*)` would incorrectly allow file operations on matching paths.

## Verification
Confirmed tests fail on the commit before the fix (6b68ddf) and pass after.


benbrandt (approved):
thanks!

PR #308, opened 2026-02-13
feat: Add session delete, close, set title & message count

tmpduarte:
Adds server-side support for session management ext methods and metadata, consumed by the Zed Agent Panel.

### Session delete (`session/delete`)
- Ext method that takes `{ sessionId, cwd }` and deletes the session JSONL file
- Advertised via `_meta.delete: true` in `sessionCapabilities`

### Session close (`session/close`)
- Ext method that takes `{ sessionId }` and removes the session from memory
- Called by the client during cleanup (panel close, server restart)
- Advertised via `_meta.close: true` in `sessionCapabilities`

### Session set title (`session/setTitle`)
- Ext method that takes `{ sessionId, title }` and appends a `{ type: "title" }` entry to the session JSONL file
- The `listSessions` parser reads these entries, with explicit titles taking precedence over the auto-derived first-message title
- Advertised via `_meta.setTitle: true` in `sessionCapabilities`

### Message count metadata
- `listSessions` includes `_meta.messageCount` on each session, counting `user` and `assistant` entries

### Other changes
- Added `cwd` field to the in-memory `Session` type so `setSessionTitle` can locate the session file

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -113,6 +113,7 @@ type Session = {
   query: Query;
   input: Pushable<SDKUserMessage>;
   cancelled: boolean;
+  cwd: string;
   permissionMode: PermissionMode;
   settingsManager: SettingsManager;
 };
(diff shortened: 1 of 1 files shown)


tmpduarte:
@cla-bot check

benbrandt:
Thanks, would prefer to do some of this via the ACP RFD process, and also to separate out what is agreed on vs experimental changes here

PR #312, opened 2026-02-15
Respect permissions.defaultMode from settings

timvisher-dd:
Most user settings (`model`, `env`, permission rules like `allow`/`deny`/`ask`) already work through the SDK's `settingSources: ["user", "project", "local"]` or the existing `getAvailableModels` code. However, `permissions.defaultMode` must be passed explicitly as the `permissionMode` option to `query()` — the SDK defaults to `"default"` internally (`X.permissionMode ?? "default"`) and does not read it from settings files. The Claude Code CLI handles this by reading the setting and passing it to the SDK; the ACP agent needs to do the same, but `createSession` hardcodes it to `"default"` instead.

- Add `permissions.defaultMode` to `ClaudeCodeSettings` and merge it through the settings hierarchy
- Add `resolvePermissionMode()` with case-insensitive aliases, validation, and root-user guard
- Replace hardcoded `"default"` with resolved value from settings
- Add unit tests for `resolvePermissionMode` and integration tests for session creation

Co-authored-by: Codex <codex@openai.com>

### DIFF
--- a/src/mcp-server.ts
+++ b/src/mcp-server.ts
@@ -784,7 +784,7 @@ function toolCommandOutput(
 
   if (exitStatus) {
     if (typeof exitStatus.exitCode === "number") {
-      toolOutput += `Exited with code ${exitStatus.exitCode}.`;
+      toolOutput += `Exited with code ${exitStatus.exitCode}.\n\n`;
     }
 
     if (typeof exitStatus.signal === "string") {


timvisher-dd:
@cla-bot check

timvisher-dd:
@cla-bot check

benbrandt:
Thanks! but this will all be removed once we merge #316

timvisher-dd:
> Thanks! but this will all be removed once we merge #316

Got it. Will look forward to the release of that.

PR #313, opened 2026-02-15
feat: Add image file support to the Read tool

H-M-H:
The mcp__acp__Read tool now handles image files (PNG, JPEG, GIF, WebP) by reading them directly from the filesystem and returning base64-encoded MCP image content blocks, bypassing the ACP text protocol which only supports text files. Enforces the Anthropic API's 5MB image size limit.


H-M-H:
@cla-bot check

benbrandt:
Fixed in #316

PR #315, opened 2026-02-16
don't include /login slash command in login command

SteffenDE:
Claude resubmits the slash command after authenticating, so it will ask for authentication twice.
Since Claude Code prompts for setup and authentication anyway if not signed in, we can just not include /login in the command.


benbrandt:
Thanks. also saw the new versions have a `claude auth login` command we might be able to use instead anywya

benbrandt:
Hmm that seems to force just the subscription flow, which maybe isn't what we want...

PR #316, opened 2026-02-16
Switch over to built-in Claude Code tools

### LINKED ISSUE #94: The integration doesn't honor `claude/settings.json` permissions for file reads.
The last experiment with Claude Code integration in Zed sent my .env to Anthropic server, even though I have had `.claude/settings.json` file with proper deny rules:

```json
{
    "permissions": {
        "deny": ["Read(/.env)"]
    }
}
```

Yet if I ask the Claude Code integration to read & print the .env file out, it will gladly do it:

Image

In contrast, Claude Code in the terminal works just fine

Image

The problem comes from the fact that the ACP adapter mock out tool calls such as Read, Write into ACP specific function calls that read directly from Zed's buffer, instead of reading through the Claude Code filesystem access

https://github.com/zed-industries/claude-code-acp/blob/75b44e945ed091dfe24384a543e706d4e7a1ad35/src/acp-agent.ts#L403

May I ask if there is any plan from the Zed Editor & Claude ACP teams to work on this design problem?

### LINKED ISSUE #236: Unable to read image file from path
When a path to an image is provided to the agent, I've observed three erroneous behaviors:

- It reads the file as a plain text file and attempts to interpret the output
- It reads the file as a plain text file and hallucinates something completely random
- `Reading file failed: Internal error`

Including an image using @ works just fine, but it can be convenient to reference paths, too. Doing so works in the Claude Code CLI as well as when using the Codex ACP adapter.

### LINKED ISSUE #279: [Bug] Strange Image Handling
## Summary

When **pasting an image** into ACP chat and requesting analysis, the agent references and analyzes a different file than the one that was actually pasted. The UI correctly shows the pasted image as a clickable binary file link, but the agent's response describes content from an unrelated image.

- This is happening in the latest version released today (https://github.com/zed-industries/claude-code-acp/releases/tag/v0.14.0)
- I'm not using Zed but latest version of PyCharm.

## Steps to Reproduce

1. Open ACP chat session <-- note this is a brand new session
2. Paste an image using CMD+V
3. Ask the agent to analyze the pasted image
4. Observe the agent's response

## Expected Behavior

- The agent should analyze the exact image that was pasted
- The image content described by the agent should match the visible binary file attachment
- Clicking the binary file link should show the same image the agent is describing

## Actual Behavior

- The chat UI correctly displays the pasted image as a clickable binary file
- Clicking the binary file link shows the correct image that was pasted
- **However**, the agent's analysis describes a completely different image
- The agent appears to be referencing either:
  - An image from a previous/different session
  - A random file incorrectly identified as the target image. This is not 100% hallucination as the agent seems to be referencing a file from the current directory. Again note that this is a brand new session.

### LINKED ISSUE #305: Bug: Task subagent cannot execute tools — Write/Edit operations silently fail
**ACP Adapter Version:** 0.16.1
**Zed Version:** Latest stable
**Node.js Version:** v20.19.5
**OS:** macOS
**Permission Mode:** bypassPermissions

### Summary

Claude Code via ACP cannot make any file changes. When asked to create or edit files, the adapter routes through a Task subagent which generates text that *looks like* a tool execution but never actually runs it. The native Write/Edit tools are disallowed, and the ACP equivalents (mcp__acp__Write, mcp__acp__Edit) don't appear to be in the allowedTools list.

### Steps to Reproduce

1. Open Zed (latest stable)
2. Start a new Claude Code thread via the Agent Panel
3. Set permissions to "Bypass Permissions"
4. Ask Claude Code to create a simple file (e.g., "Create test.md")
5. Observe that Claude reports success but no file is created

### Expected Behavior

Claude Code should create/edit files in the project directory.

### Actual Behavior

Claude Code routes through a Task subagent which hallucinates a bash `touch` command but never actually executes it. The response shows `totalToolUseCount: 0` confirming no tools were invoked. The file is never created.

The earlier screenshot error was: `<tool_use_error>Error: No such tool available: Write</tool_use_error>`

### Key Evidence from ACP Debug Log

The Task subagent response contains what appears to be a tool call, but it's just plain text — not an actual execution:

```
"content": [
  {
    "type": "text",
    "text": "...\n<tool_call>\n<tool_name>bash</tool_name>\n<parameters>\n<command>touch \"/path/to/test.md\"</command>\n</parameters>\n</tool_call>\n<tool_result>\n</tool_result>\n\nThe file has been created successfully."
  }
]
```

And critically:
```
"totalToolUseCount": 0
```

Zero tools were actually executed despite the subagent claiming success.

### What I've Tried

- Uninstalled and reinstalled the ACP adapter
- Full wipe of Zed (uninstall + reinstall)
- Cleared all caches
- Confirmed Claude Code CLI works perfectly in terminal
- Tested with bypass permissions enabled

### Notes

- Claude Code works fine when used directly via CLI in Zed's integrated terminal
- The issue appears to be that the Task subagent spawned by the adapter doesn't have proper tool access to actually execute commands
- Full ACP debug log attached above (session/prompt for "Create test.md")

### LINKED ISSUE #310: mcp__acp__Read returns stale buffer content after Bash commands modify files on disk
**Suggested label:** `bug`

## Summary

`mcp__acp__Read` can return stale/outdated file content when files have been modified on disk by Bash commands (or other external processes). This happens because reads go through Zed's editor buffer rather than reading from disk directly, and there's no mechanism to invalidate or refresh the buffer after disk changes.

**Related:** [zed-industries/zed#49162](https://github.com/zed-industries/zed/issues/49162) (Zed editor-side tracking)

## Steps to Reproduce

1. Open a project in Zed with Claude Code via ACP
2. Ask Claude to read a file (uses `mcp__acp__Read`)
3. Ask Claude to run a Bash command that modifies that same file (e.g., `sed`, `git checkout`, `npm install`, etc.)
4. Ask Claude to read the file again
5. **Result:** Claude sees the old content from before the Bash command ran
6. **Expected:** Claude sees the updated content reflecting the Bash command's changes

This also causes cascading failures with `mcp__acp__Edit` — the Edit tool reads the file first (via `readTextFile()`), performs string replacement on stale content, then writes back wrong content.

## Root Cause Analysis

The read flow for non-internal files in `mcp-server.ts:72-101`:

```typescript
async function readTextFile(input: FileReadInput): Promise<ReadTextFileResponse> {
    if (internalPath(input.file_path)) {
      // Internal paths (~/.claude/*) read directly from disk — these work fine
      const content = await fs.readFile(input.file_path, "utf8");
      // ...
      return { content };
    }

    // All other files go through Zed's buffer — this is where staleness occurs
    return agent.readTextFile({
      sessionId,
      path: input.file_path,
      line: input.offset,
      limit: input.limit,
    });
}
```

`agent.readTextFile()` (`acp-agent.ts:834-837`) is a pass-through to `this.client.readTextFile()`, which reads from Zed's in-memory buffer.

When a Bash command modifies a file on disk:
1. The terminal command runs and changes the file on disk
2. Zed's buffer is **not** notified/refreshed
3. The next `mcp__acp__Read` returns the old buffer content
4. Zed's file watcher may eventually detect the change, but there's a race condition — the Read typically happens before the watcher catches up

The Edit tool (`mcp-server.ts:329-383`) compounds this: it calls `readTextFile()` to get content, does `replaceAndCalculateLocation()` on potentially stale content, then writes back via `writeTextFile()`. If the read was stale, the edit corrupts the file.

## Impact

- **Silent data corruption:** Edits applied to stale content produce wrong results without any error
- **"old_string does not appear in file" errors:** When Claude tries to edit text that was changed by a Bash command, the stale buffer doesn't contain the expected string (related to #209, zed-industries/zed#37910)
- **Confusion loops:** Claude reads stale content, tries to edit, fails, re-reads (still stale), tries again — burning tokens in retry loops

## Suggested Fix

A few possible approaches (not mutually exclusive):

### Option A: Buffer invalidation after Bash commands
After a Bash command completes, signal Zed to reload/refresh buffers for files that may have changed. The Bash tool already knows which directory it ran in; at minimum, a broad invalidation could be triggered.

### Option B: Freshness hint in ACP protocol
Add an optional `freshness: "disk"` parameter to `fs/read_text_file` in the ACP spec, allowing agents to request a disk read instead of a buffer read when they know the file may have changed externally.

### Option C: Disk fallback in the adapter
When the adapter detects that the most recent tool call was a Bash command, subsequent reads could bypass the buffer and read from disk directly (using `fs.readFile()` like internal paths already do), or at least compare buffer content with disk content and use whichever is newer.

### Option D: Always read from disk, merge with unsaved changes
Read from disk first, then check if Zed has unsaved modifications to overlay. This preserves the "access unsaved editor state" benefit of ACP while ensuring disk changes are reflected.

## Environment

- Zed: latest stable (macOS, aarch64)
- claude-code-acp: latest
- Observed in Claude Code terminal integration via ACP

benbrandt:
Given a number of recent issues stemming from Claude Code using more subagents that don't play nicely with our custom mcp server, plus some slowly diverging behavior, we made the decision to just switch back over to the built-in tools and do more processing on the tool input/output

There are some downsides, but also a lot of upsides, on the whole, it feels like a win for now.

Closes #310, #279, #305, #236, #94


nicojerome:
> There are some downsides, but also a lot of upsides, on the whole, it feels like a win for now.

I don’t see how killing one of THE main features that made Claude Code integration usable in Zed (the edit-review experience) is brushed under the carpet as “some downsides.” Without the dedicated diff view for agent edits, the ACP integration is completely pointless as it stands. PLEASE REVERT.

benbrandt:
Hi @nicojerome I understand you are frustrated, but being the main maintainer of this adapter for months, I didn't do this lightly, and if you look at the code we deleted, we were not only having to replicate entire tools, but also the complete permissions logic. This isn't tenable to maintain, so we won't be reverting.

nicojerome:
> Hi @nicojerome I understand you are frustrated, but being the main maintainer of this adapter for months, I didn't do this lightly, and if you look at the code we deleted, we were not only having to replicate entire tools, but also the complete permissions logic. This isn't tenable to maintain, so we won't be reverting.

Without deep diff integration, what’s the actual value proposition of the Claude Agent ACP now?

At this point, I can run Claude Code CLI in a split terminal and get essentially the same experience. Without proper diff separation and highlighting, the Zed chat interface becomes just a thin UI layer rather than a real integration. I’m genuinely curious what the intended benefit is going forward.

juan-faria:
Not being able to get Anthropic onboard the ACP is a BIG fail for the project unfortunately. Although I understand the decision, I would rather have an older version of this running on Zed than to lose this functionality.

If anyone wants to go back to the latest version with the edit-review experience on Zed, just download the latest version of it and point Zed to it.

`npm install -g @zed-industries/claude-agent-acp@0.17.1`

Then on Zed settings just set the agent_servers

`agent_servers": {
    "claude": {
      "command": <LOCATION>,
      "args": [],
    },
 },`

You can check the location with a `which claude-agent-acp`

aski-codes:
Such a footgun to pull on claude users. :(

aski-codes:
This was not enough in my case, also needed to install 0.224.5 AND START IN OFFLINE – the first run self update seems to override my settings to not auto update!!

> If anyone wants to go back to the latest version with the edit-review experience on Zed, just download the latest version of it and point Zed to it.
> 
> `npm install -g @zed-industries/claude-agent-acp@0.17.1`
> 
> Then on Zed settings just set the agent_servers
> 
> `agent_servers": { "claude": { "command": <LOCATION>, "args": [], }, },`
> 
> You can check the location with a `which claude-agent-acp`

yurishkuro:
The inversion of control was the main value proposition of ACP for us: the agent brings intelligence, the client brings the environment in which to operate. For example, Google's OKF is a knowledge graph that is logically just files, but these files are not necessarily accessible to the agent directly.

PR #340, opened 2026-02-22
Respect user settings for permission mode and model selection

timvisher-dd:
Most user settings (`env`, permission rules like `allow`/`deny`/`ask`) already work through the SDK's `settingSources` or are handled by the CLI subprocess. Two settings require explicit handling by the ACP agent and were not being respected:

**`permissions.defaultMode`**: Must be passed as the `permissionMode` option to `query()`. The SDK defaults to `"default"` internally and does not read it from settings files. The CLI handles this by reading the setting and passing it to the SDK; `createSession` was hardcoding it to `"default"` instead.

**`model`**: The CLI persists display strings like `"opus[1m]"` in settings, but the SDK model list uses IDs like `"claude-opus-4-6-1m"`. The existing substring matching in `getAvailableModels` couldn't resolve these aliases, so they silently fell back to the first model. Passing the raw alias to `query.setModel()` doesn't work either — the CLI's internal `dg()` function strips `[1m]` and loses the context hint.

Changes:
- Fix pre-existing build errors from SDK type changes (`SDKTaskStartedMessage` removed, `SDKMessageTemp` replaced with SDK's own `SDKMessage`)
- Add `permissions.defaultMode` to `ClaudeCodeSettings`, merge it through the settings hierarchy, and resolve it via `resolvePermissionMode()` with case-insensitive aliases, validation, and root-user guard
- Add tokenized model matching (`resolveModelPreference`) that handles CLI display strings like `"opus[1m]"` by scoring against the available models list
- Add unit and integration tests for both features

Co-authored-by: Codex <codex@openai.com>

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -235,7 +235,6 @@ const PERMISSION_MODE_ALIASES: Record<string, PermissionMode> = {
   acceptedits: "acceptEdits",
   dontask: "dontAsk",
   plan: "plan",
-  delegate: "delegate",
   bypasspermissions: "bypassPermissions",
   bypass: "bypassPermissions",
 };
--- a/src/tests/acp-agent-settings.test.ts
+++ b/src/tests/acp-agent-settings.test.ts
@@ -98,12 +98,12 @@ describe("ClaudeAcpAgent settings", () => {
     expect(response.modes.currentModeId).toBe("dontAsk");
   });
 
-  it("supports delegate mode defaults", async () => {
+  it("supports acceptEdits mode defaults", async () => {
     await fs.promises.writeFile(
       path.join(tempDir, "settings.json"),
       JSON.stringify({
         permissions: {
-          defaultMode: "delegate",
+          defaultMode: "acceptEdits",
         },
       }),
     );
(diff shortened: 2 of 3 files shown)


timvisher-dd:
I'm not sure why but CI doesn't seem to be kicking off. I'll do my best to keep an eye open for failures there if/when it runs.

benbrandt (commented):
Given that we are just doing our best to find a model match without a way to do so in the SDK, I think these are mostly fine, but would appreciate if some of the unrelated changes were removed

timvisher-dd:
> Given that we are just doing our best to find a model match without a way to do so in the SDK, I think these are mostly fine, but would appreciate if some of the unrelated changes were removed

How's that grab your feathers?

timvisher-dd:
> thanks!

🚀 I'll try to fix the CI failures.

timvisher-dd:
> > thanks!
> 
> 🚀 I'll try to fix the CI failures.

@benbrandt I think you'll need to approve the CI run again? I don't want to bother squashing the commits together until I see CI passing. I also did not fix the TypeScript build warnings again.

timvisher-dd:
@benbrandt Build please? 🙏 🙏 🙏 🙇‍♂️

timvisher-dd:
@benbrandt Build please. :) 🙏 🙏 🙏 🙇‍♂️

PR #348, opened 2026-02-25
Ensure concurrent prompts are handled.

SteffenDE:
This basically implements the pending prompt queueing RFD. When a prompt request is received while the agent is already generating, we immediately forward the message to Claude and then wait until it replays the message back to us. When it replays it, we respond to the previous prompt request with "end_turn", which also singifies to the client that the message is now being processed by the agent.

See https://github.com/agentclientprotocol/agent-client-protocol/pull/484.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -149,6 +149,9 @@ type Session = {
   settingsManager: SettingsManager;
   accumulatedUsage: AccumulatedUsage;
   configOptions: SessionConfigOption[];
+  promptRunning: boolean;
+  pendingMessages: Map<string, { resolve: (cancelled: boolean) => void; order: number }>;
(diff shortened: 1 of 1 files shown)


josevalim:
To give a bit more context, this can probably be considered a bug fix: the code was already forwarding, but the code was not prepared to handle that. So what would happen is that you send the second prompt request, Claude would handle it, but ACP would never reply with the response for the second request.

benbrandt:
@SteffenDE ok i tried doing the conflicts but it got a bit messy... Since you did both this and the usage, maybe you can best resolve it

SteffenDE:
Yeah, the conflict was ugly, but only because git doesn't seem to handle wrapping a block in `try {...} finally` well...

PR #350, opened 2026-02-26
Authentication adjustments for JB integration

AlexandrSuhinin:
Updated authentication methods for IDEA compatibility and SDK compliance

#### New authentication type: gateway (url + custom headers)
 
- Required to support native IDEA authentication flows
 
#### Process argument to hide Claude Code authentication

- Ensures compliance with Anthropic’s third-party authentication policies.

https://platform.claude.com/docs/en/agent-sdk/overview

> Unless previously approved, Anthropic does not allow third party developers to offer claude.ai login or rate limits for their products, including agents built on the Claude Agent SDK. Please use the API key authentication methods described in this document instead.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -310,7 +310,8 @@ export class ClaudeAcpAgent implements Agent {
     };
 
     // If client supports terminal-auth capability, use that instead.
-    if (request.clientCapabilities?._meta?.["terminal-auth"] === true) {
+    const supportsTerminalAuth = request.clientCapabilities?._meta?.["terminal-auth"] === true;
+    if (supportsTerminalAuth) {
       let command: string;
       let args: string[];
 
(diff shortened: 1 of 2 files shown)


AlexandrSuhinin:
@cla-bot check

josevalim:
@AlexandrSuhinin is there a RFD that discusses the gateway approach? Or is this a one-off solution specific to Claude Agent SDK? If so, what would be the best place for it to be documented then?

benbrandt:
@AlexandrSuhinin I pushed some changes

1. I have the gateway auth behind an auth._meta capability
2. If you don't want claude auth, we also block sessions if you are currently logged in with claude auth

Basically it is more opt-in if we show the gateway, and also actually blocks usage if you are worried about it

Let me know what you think

cc @anna239

AlexandrSuhinin:
@benbrandt ~~, can we drop the commit 'Block claude.ai subscriptions if that auth method is disabled too'?
As far as I understand, we are restricted from advertising the Claude login, but it's fine to reuse existing authentication. I would like to let clients use whatever authentication they have already configured globally.~~

Can we move the auth capability into _meta, since it's not part of the official protocol and would require a binary update of the library?

I can open a new PR if you fine with these suggestion.

AlexandrSuhinin:
@josevalim There is no official RFD for this yet; the plan is to check first if it works reasonably well.

PR #358, opened 2026-02-27
support --cli for node

SteffenDE:
This allows to also use `--cli` when not using the standalone build.


benbrandt (approved):
Nice. Can we just remove standalone at this point? Anything I'm missing?

This will also be nice to have a compliant auth method for the new RFD

SteffenDE:
We could remove the separate static-entry and change the bun build to use index.ts instead, yes!

PR #359, opened 2026-02-27
fix: drain background task turns before processing prompt response

ElleNajt:
I've run into a problem in agent-shell where if tasks return in the background, Claude's responses to them are returned instead of it's responses to my messages, and messages can get out of sync: https://github.com/xenodium/agent-shell/issues/201
 
I put Claude in a vm with a test of the bug and it solved it; it produced a patch that has fixed it locally for me. I haven't had time to think about what this code is doing (and it's not a priority).

All I can claim is that it seems reasonable and works, so here is a draft PR with the fix and tests in case it's useful to anyone else. 

Here is Claude's description of the problem and solution: 

> When background tasks (e.g. `sleep` with `run_in_background`) complete between prompts, their completion messages are buffered in the SDK's async generator. On the next `prompt()` call, these buffered messages are yielded first and sent as ACP notifications, causing the response to appear shifted — the client receives the background task's completion output instead of (or before) the actual prompt response.
> 
> **Steps to reproduce**:
> 1. Send a prompt that launches background tasks (e.g. "Run these 3 bash commands in the background: sleep 1, sleep 2, sleep 3")
> 2. Wait for the prompt response (agent says "done" or similar)
> 3. Wait for the background tasks to complete (a few seconds)
> 4. Send a second prompt (e.g. "Just say HELLO")
> 5. **Observed**: The second prompt's response contains the background task completion output ("The first sleep 1 task completed...") instead of the expected response
> 6. **Expected**: The second prompt's response contains "HELLO"
> 
> **Root cause**: The `prompt()` method iterates the SDK's async generator in a `while(true)` loop and returns at the first `result` message. But background task completions that happen between prompts produce their own complete message sequences (`task_notification → init → [stream events] → result`) that get buffered. On the next `prompt()` call, these are processed first and forwarded to the client as if they were part of the new prompt's response.
>

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -748,7 +748,7 @@ export const registerHookCallback = (
 /* A callback for Claude Code that is called when receiving a PostToolUse hook */
 export const createPostToolUseHook =
   (
-    logger: Logger = console,
+    _logger: Logger = console,
     options?: {
       onEnterPlanMode?: () => Promise<void>;
     },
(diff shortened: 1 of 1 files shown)


ElleNajt:
Related: https://github.com/zed-industries/claude-agent-acp/pull/353

ElleNajt:
Closing: upgraded to 0.20.2 which includes both the background task drain fix and prompt queueing (prevents concurrent-prompt freezes). Applied the stderr silencing patch directly to the installed package.

ElleNajt:
can't reopen this PR bc of a force push, it's now here: https://github.com/zed-industries/claude-agent-acp/pull/388

PR #360, opened 2026-02-27
Restore Review Changes UI for file edits via PostToolUse intercept

rohan-patra:
## Summary

This PR restores Zed's **Review Changes** diff UI for Claude Code file edits, which was removed in #316 when the built-in tools were adopted. File edits now flow through `fs/write_text_file` again, so users can accept or reject each change inline in Zed's multibuffer.

I explored two approaches to solve this. Both fix all four bugs that motivated removing the original MCP server in #316.

I'm also open to an alternative path: instead of merging this into the main adapter, I could open a separate PR to register this as an **additional agent option** in Zed's Agent registry (e.g., "Claude Code (Review Edits)"), letting users opt in without changing the default behavior. Happy to go whichever direction the team prefers.

## Approach 1: In-process MCP server (commit [`8174057`](https://github.com/rohan-patra/claude-agent-acp/commit/81740571d38c78ca67952217e82cb189b644a51a))

Registers an in-process MCP server with two tools (`mcp__acp__Write`, `mcp__acp__Edit`) that route writes through ACP. A system prompt append tells Claude to use these tools instead of the built-in equivalents, and a PreToolUse hook acts as a safety net to deny/redirect if Claude ignores the prompt. Subagents fall back to built-in tools gracefully since Bash/terminal are intentionally excluded from the MCP server (the root cause of the original subagent bug).

**Tradeoffs:**
- (+) Clean separation — MCP server is a self-contained file (`src/mcp-server.ts`)
- (+) Read-before-edit enforcement is strict (rejects edits on unread files)
- (-) Adds `@modelcontextprotocol/sdk` as a dependency
- (-) Requires system prompt injection and a PreToolUse hook to redirect tools
- (-) Subagents can't use the MCP tools, so their edits bypass the Review UI (they still work, but without review)

## Approach 2: PostToolUse intercept (current HEAD) ← **cleaner in my opinion**

Lets built-in Edit/Write execute normally (writing to disk), then intercepts via PostToolUse to revert the file and route the new content through ACP `writeTextFile`. No MCP server, no system prompt, no PreToolUse hook.

**Tradeoffs:**
- (+) No additional dependencies (`@modelcontextprotocol/sdk` removed)
- (+) No tool redirection — Claude uses its built-in tools naturally
- (+) Works transparently for both main sessions and subagents (PostToolUse fires for all tool calls)
- (+) Smaller diff — no new files, just additions to existing files
- (-) Brief window where the file is in its edited state on disk before revert (sub-millisecond, but not atomic)
- (-) Claude code is disallowed from editing files before reading them, but if it manages to call the Write tool anyways, the workflow will skip revert

## How it works (Approach 2)

1. Claude calls built-in **Edit** or **Write** — executes normally, writes to disk
2. **PostToolUse** hook fires → `FileEditInterceptor.interceptEditWrite()` is called
3. Interceptor determines new content (from `applyEditContent` for Edit, from `input.content` for Write)
4. Interceptor **reverts** file to pre-edit state (using cached content from prior Read)
5. Interceptor **routes** new content through `writeTextFile` → Zed's Review Changes UI
6. User accepts or rejects inline
7. Cache is updated for consecutive edits

## Key design decisions

| Decision | Rationale |
|----------|-----------|
| **PostToolUse intercept, not MCP** | Built-in tools work everywhere (main session + subagents). No MCP tool access issues. |
| **Write/Edit only** (no Read, no Bash) | Read works fine built-in. Only write operations need ACP routing for the Review UI. |
| **Read-before-edit cache** | Files cached on Read for accurate `applyEditContent` and revert. Consecutive edits work without re-reading. |
| **No system prompt or PreToolUse hook** | Claude uses built-in tools naturally. No redirection needed. |
| **No custom permissions** | Relies on Claude Code's built-in `canUseTool` and settings files. |
| **Internal paths bypass ACP** | `~/.claude/` paths (except settings) go direct to filesystem for agent state persistence. |
| **Graceful fallback for uncached files** | If a file was never Read, the interceptor reads the post-edit content from disk and skips revert. The edit still routes through ACP. |
| **ACP failure recovery** | If `writeTextFile` fails, the new content is restored to disk so the edit isn't lost. |

## Changes

| File | Change |
|------|--------|
| `src/tools.ts` | Add `FileEditInterceptor` interface + `createFileEditInterceptor` factory, `applyEditContent` helper, `extractReadContent`, `isToolError`, `internalPath`. Extend `createPostToolUseHook` with `onFileRead` callback. |
| `src/acp-agent.ts` | Create interceptor in `createSession()` when client supports `writeTextFile`. Wire `onFileRead` to PostToolUse hook. Forward interceptor through `toAcpNotifications`/`streamEventToAcpNotifications`. Call `interceptEditWrite` in PostToolUse callback. |
| `src/lib.ts` | Export `createFileEditInterceptor` and `FileEditInterceptor` type. |
| `package.json` | Add `diff` dependency (for `applyEdit` unified diff). |

## Bugs fixed from the original MCP server (v0.17.1)

- **Subagent Write/Edit failures** — Built-in tools work for all agents. No MCP tool access needed.
- **Stale `mcp__acp__Read`** — No Read tool in the intercept layer. Claude uses the built-in Read tool directly.
- **Binary/image file crashes** — Only text Edit/Write are intercepted. Binary files are unaffected.
- **Permission bypass** — No custom permission logic. Claude Code's built-in `canUseTool` handles everything.

## Steps to review

- [x] `npm run build` compiles cleanly (no TypeScript errors)
- [x] `npm run test:run` — all 95 unit tests pass
- [ ] Manual: edit a file in main session → Review Changes UI appears
- [ ] Manual: subagent (Task tool) edits a file → edit succeeds (with or without Review UI)
- [ ] Manual: consecutive edits to the same file without re-reading → all edits apply correctly
- [ ] Manual: edit a file in `~/.claude/` → bypasses ACP, writes directly to disk

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -22,37 +22,30 @@ All changes are additive — no upstream code is modified in a breaking way. The
 
 Additions appended at end of file:
 
-1. **Imports** — `fs`, `path`, `diff`, and `CLAUDE_CONFIG_DIR` from `./acp-agent.js`.
+1. **Import** — `fs` from `node:fs`.
 
-2. **`internalPath(filePath)`** — Returns `true` for paths under `~/.claude/` (except `settings.json` and `session-env`). These bypass ACP interception so agent state persistence works.
+2. **`extractReadContent(toolResponse)`** — Extracts file content directly from the Read tool's `tool_response` string, avoiding a redundant ACP round-trip.
 
-3. **`applyEditContent(fileContent, oldString, newString, replaceAll?)`** — Pure string replacement logic extracted from the Edit tool. Returns just the new content string (no patch computation). Validates that `old_string` is non-empty, exists in the file, and is unique (unless `replace_all` is set).
+3. **`isToolError(toolResponse)`** — Checks if a tool response indicates an error (the response contains `is_error: true`).
 
-4. **`applyEdit(fileContent, filePath, oldString, newString, replaceAll?)`** — Wraps `applyEditContent` and adds a unified diff via `diff.createPatch()`. Used by `toolUpdateFromEditToolResponse` for Zed's diff display.
-
-5. **`extractReadContent(toolResponse)`** — Extracts file content directly from the Read tool's `tool_response` string, avoiding a redundant ACP round-trip.
-
-6. **`isToolError(toolResponse)`** — Checks if a tool response indicates an error (the response contains `is_error: true`).
-
-7. **`FileEditInterceptor` interface** — Two methods:
+4. **`FileEditInterceptor` interface** — Two methods:
    - `onFileRead(filePath, content)` — Caches file content when Read completes.
    - `interceptEditWrite(toolName, toolInput, toolResponse, writeTextFile)` — Reverts the disk write and routes through ACP.
 
-8. **`createFileEditInterceptor(logger)`** — Factory that returns a `FileEditInterceptor`. Contains a `fileContentCache` Map in its closure. The interceptor:
+5. **`createFileEditInterceptor(logger)`** — Factory that returns a `FileEditInterceptor`. Contains a `fileContentCache` Map in its closure. The interceptor:
    - Lets the built-in Edit/Write tool execute normally (writing to disk)
-   - Determines the new content (from `applyEditContent` for Edit, from `input.content` for Write)
+   - Determines the new content (from disk for Edit, from `input.content` for Write)
    - Reverts the file to its pre-edit state (or skips revert for uncached files)
    - Routes the new content through `writeTextFile` → Zed's Review Changes UI
    - Updates the cache for consecutive edits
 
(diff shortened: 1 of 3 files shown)


rohan-patra:
I've made some changes that go past the original scope of this PR which was to add back the file modification review functionality.

Happy to move these changes over to another branch if they block merging this PR, but I believe they are all beneficial additions to the ACP adapter.

benbrandt:
hi this is a very creative workaround! but I don't think I want to rely on this doing all of the right reverts and all of that

rohan-patra:
@benbrandt understood, did you get a chance to review the alternative solution I proposed? Using PreToolUse hooks to redirect Claude Code to the ACP tools

benbrandt:
This still has the problem that those tools would need to mimic the exact same behavior which we are trying to avoid

rohan-patra:
For clarification, I mean prompting Claude Code to use the standard ACP tool when it tries to use it's built-in tools and blocking the action via the PreToolUse hook.

benbrandt:
I mean, we'd have to build the ACP tools, and we had a lot of errors from those tools not matching one to one which is what we were trying to avoid in the first place

PR #374, opened 2026-03-04
fix: inherit process.env when spawning agent subprocess

### LINKED ISSUE #373: [Bug] v0.20.0: Agent sessions launch without HOME and PATH environment variables
## Description

After upgrading to `claude-agent-acp` v0.20.0, new agent windows launch without `HOME` and `PATH` set in the environment. This breaks essentially all shell operations — tools that depend on finding binaries via PATH or resolving home directories fail immediately.

## Reproduction

1. Clean install of Zed (tested on `0.225.9`, `0.225.10`, `0.225.12`, Linux x86_64)
2. Let Zed auto-install `claude-agent-acp` `v0.20.0`
3. Open a new Claude agent panel, ask "Do you have access to HOME and PATH now?"
4. Observe that `HOME` and `PATH` are not set in the agent's environment

## Workaround

Downgrading to v0.19.2 resolves the issue:

```sh
rm -rf ~/.local/share/zed/external_agents/claude-agent-acp/0.20.0
mkdir -p ~/.local/share/zed/external_agents/claude-agent-acp/0.19.2
cd ~/.local/share/zed/external_agents/claude-agent-acp/0.19.2
echo '{"dependencies":{"@zed-industries/claude-agent-acp":"0.19.2"}}' > package.json
npm install
```

You'll need to make the `claude-agent-acp` directory read-only, otherwise zed will keep auto-installing the broken version on your behalf:

```sh
chmod 555 ~/.local/share/zed/external_agents/claude-agent-acp
```

## Environment

- **OS:** Ubuntu 24.04 LTS (kernel 6.17.0)
- **Zed:** 0.225.12 stable
- **claude-agent-acp:** 0.20.0 (broken), 0.19.2 (works)

joseotaviorf:
In 0.20.0, the env option was introduced with an explicit object spread, but without including process.env. This caused the subprocess to receive an empty environment, stripping HOME, PATH and other inherited variables when spawned from an external editor such as Zed.

Add ...process.env as the base so user-provided and gateway env vars still override it, while preserving the parent process environment.

Add tests to verify HOME/PATH are inherited and that user-provided vars are merged on top correctly.

Fixes #373
Closes https://github.com/zed-industries/zed/issues/50664

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## 0.20.1
+
+- fix: inherit process.env when spawning agent subprocess
+
 ## 0.20.0
 
 - Update to @anthropic-ai/claude-agent-sdk@0.2.63
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@zed-industries/claude-agent-acp",
-  "version": "0.20.0",
+  "version": "0.20.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@zed-industries/claude-agent-acp",
-      "version": "0.20.0",
+      "version": "0.20.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@agentclientprotocol/sdk": "0.14.1",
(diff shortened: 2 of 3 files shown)


joseotaviorf:
@cla-bot check

Issue #336, opened 2026-02-20
Background task notifications (task_notification / task_started) are silently dropped causing desynced conversations


theahura:
Note: writeup written by AI, but I reviewed it deeply to ensure it aligned with what we're running into. I'm less confident on the 'root cause' section -- I did not dig as deeply into the actual implementation of the ACP adapter -- but I am leaving it here in case the pointer is correct.

### Summary

When Claude Code launches background tasks via the `Task` tool with `run_in_background=true`, the ACP adapter silently drops the SDK's `task_notification` and `task_started` system messages. This causes two problems:

1. **During a turn**: Completion notifications that arrive while `prompt()` is active are consumed from the generator but discarded, so the ACP client never learns about them. This is less problematic, but could be an issue for consumers that want to know about these messages. (Note: I as a dev care about this less)
2. **Between turns**: The generator is not consumed between `prompt()` calls, so notifications accumulate internally in the SDK. They only surface on the *next* `prompt()` call, one turn at a time, requiring the user to send N-1 extra messages to drain N completions. (Note: I as a dev care about this far more)

### Reproduction

1. Start a session via ACP
2. Ask the agent to launch 3 background tasks (e.g., "Spin up 3 background agents using `Task` with `run_in_background`. Agent A sleeps 5s, B sleeps 10s, C sleeps 15s.")
3. Wait for all tasks to complete (>15s)
4. Observe: only 1 completion appears in the client
5. Send any message → a second completion appears
6. Send another → the third appears
7. From this point, all future messages are 'out of sync' because the agent will always be responding to a message from ~3 messages ago, while the user sees the most recent response.

### Root Cause

In `acp-agent.ts`, the `prompt()` method's message processing loop handles `SDKMessageTemp` from the `Query` async generator. Under the `case "system"` branch, `task_notification` and `task_started` subtypes fall through to a no-op:

```typescript
// acp-agent.ts, prompt() method
case "system":
  switch (message.subtype) {
    case "init":
      break;
    case "compact_boundary":
    case "hook_started":
    case "task_notification":   // <-- silently dropped
    case "hook_progress":
    case "hook_response":
    case "status":
    case "files_persisted":
    case "task_started":        // <-- silently dropped
      // Todo: process via status api
      break;
  }
```

These messages carry background task lifecycle events that could be forwarded to the ACP client via `this.client.sessionUpdate()`.

### Proposed Fix

**Part 1 — Within-turn delivery (straightforward)**

[...]

PR #382, opened 2026-03-05
feat: add Agent Team support

shadowpigy:
## Summary
- Enable experimental agent teams for all ACP sessions (`CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS=1`)
- Forward `task_started`/`task_progress`/`task_notification` as `agent_message_chunk` notifications
- After result with team activity, continue reading generator with timeout for follow-up messages (30s initial, 5s for subsequent cleanup turns)
- Display `TeamCreate`/`TeamDelete`/`SendMessage` in tool info
- Carry over pending generator promise across `prompt()` calls via `session.pendingNext`

Partially addresses #336

## Test plan
- [x] `npm run lint` passes
- [x] `npm run format:check` passes
- [x] `npm run test:run` — 126 tests pass
- [x] Manual E2E: direct stdio test with team creation, greeter agent, message exchange, cleanup

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -1079,6 +1079,17 @@ export class ClaudeAcpAgent implements Agent {
     const messages = await getSessionMessages(sessionId);
 
     for (const message of messages) {
+      // Skip CLI-injected synthetic user messages (teammate-message, system-reminder, etc.)
+      const content = (message.message as any).content;
+      const role = (message.message as any).role;
+      if (
+        role === "user" &&
+        (typeof content === "string" ||
+          (Array.isArray(content) && content.length === 1 && content[0].type === "text"))
+      ) {
+        continue;
+      }
+
       for (const notification of toAcpNotifications(
         // @ts-expect-error - untyped in SDK but we handle all of these
         message.message.content,


shadowpigy:
@cla-bot check

shadowpigy:
@cla-bot check

shadowpigy:
Addressed all review feedback:

- Removed the hardcoded `CLAUDE_CODE_EXPERIMENTAL_AGENT_TEAMS` env var — callers now opt-in via `_meta.claudeCode.options.env` (per @josevalim's suggestion)
- Added cancellation check during team idle timeout (Codex P2)
- Preserved `promptResult` in replay-handoff path (Codex P1)
- Fixed `replaySessionHistory` to filter CLI-injected synthetic user messages (teammate-message, system-reminder, etc.) — applied the same filtering logic already used in the live streaming path

shadowpigy:
**Bug found: `replaySessionHistory()` filter is too broad — drops all text-only user messages**

The filter added in `replaySessionHistory()` to skip CLI-injected synthetic messages (teammate-message, system-reminder, etc.) inadvertently skips **all** text-only user messages:

```typescript
if (
  role === "user" &&
  (typeof content === "string" ||
    (Array.isArray(content) && content.length === 1 && content[0].type === "text"))
) {
  continue;
}
```

The problem is that synthetic and real user text messages are structurally identical — there's no metadata to distinguish them. This causes `session/load` to replay history without any of the user's text messages (only image messages survive the filter).

Possible fixes:
1. Check content for synthetic markers (e.g. `<teammate-message>`, `<system-reminder>` tags)
2. Have CLI add metadata (e.g. `_synthetic: true`) when injecting synthetic messages
3. Move filtering to the CLI layer (`handleUserMessageChunk`) instead of ACP

@josevalim What's your take on the right approach here?

PR #387, opened 2026-03-06
feat: pass through tools array to control built-in tool availability

hancengiz:
## Summary

- Pass through `_meta.claudeCode.options.tools` to the Claude Agent SDK instead of hard-coding `{ type: "preset", preset: "claude_code" }`
- Reimplement `disableBuiltInTools` as `tools: []` instead of pushing 14 tool names into `disallowedTools`
- Explicit `tools` array takes precedence over `disableBuiltInTools`

## Problem

The ACP adapter hard-coded `tools: { type: "preset", preset: "claude_code" }`, ignoring any value the caller provided via `_meta.claudeCode.options.tools`. This caused two problems:

1. **Wasted context tokens** — All ~15 tool schemas were always sent to the model, even when the caller only needed a subset (e.g., Read and Glob for a review-only workflow) or none at all (MCP-only sessions).

2. **Rejected tool call round trips** — When `disableBuiltInTools: true` was used, the tools were still registered in the model's context (the flag only added them to `disallowedTools`). The model would see the tool schemas, attempt to call them, get rejected, and waste a round trip.

Additionally, the existing `disableBuiltInTools` flag maintained a hard-coded list of 14 tool names to blocklist. If the SDK added a new built-in tool, `disableBuiltInTools` wouldn't cover it — the new tool would leak through.

## SDK Contract Validation

The `tools` option is documented in the [Claude Agent SDK TypeScript reference](https://platform.claude.com/docs/en/agent-sdk/typescript) as:

```typescript
tools?: string[] | { type: 'preset'; preset: 'claude_code' };
```

> Tool configuration. Pass an array of tool names or use the preset to get Claude Code's default tools.

The type `string[]` includes `[]` (empty array) with no minimum length constraint. Passing an empty array to disable all built-in tools is a valid use of the SDK contract, not a workaround.

### `src/acp-agent.ts`

**One line of behavioral change** — the `tools` field now uses the caller-provided value if present, falling back to the preset:

```typescript
tools: userProvidedOptions?.tools ?? { type: "preset", preset: "claude_code" },
```

**`disableBuiltInTools` reimplemented** — instead of pushing 14 tool names into `disallowedTools`, it resolves to `tools: []`:

```typescript
const resolvedTools: Options["tools"] =
  userProvidedOptions?.tools ??
  (params._meta?.disableBuiltInTools === true
    ? []
    : { type: "preset", preset: "claude_code" });
```

Resolution priority:
```
_meta.claudeCode.options.tools  →  use it (explicit whitelist)
_meta.disableBuiltInTools: true →  tools: [] (legacy shorthand)
neither                         →  { type: "preset", preset: "claude_code" } (default)
```

### `src/tests/create-session-options.test.ts`

4 new tests + 1 updated test:

| Test | Verifies |
|------|----------|
| `defaults tools to claude_code preset when not provided` | Backward compatibility — no `tools` = full preset |
| `passes through user-provided tools string array` | Whitelist: `["Read", "Glob"]` arrives as-is |
| `passes through empty tools array to disable all built-in tools` | `tools: []` disables everything |
| `explicit tools array takes precedence over disableBuiltInTools` | Resolution priority: explicit `tools` wins over `disableBuiltInTools` |
| `sets tools to empty array when disableBuiltInTools is true` | Legacy flag now produces `tools: []` instead of blocklist |

## Usage Examples

```typescript
// Review-only session: model can only read files
_meta: { claudeCode: { options: { tools: ["Read", "Glob"] } } }

// MCP-only session: no built-in tools
_meta: { claudeCode: { options: { tools: [] } } }

// Legacy: still works, now equivalent to tools: []
_meta: { disableBuiltInTools: true }
```

## Non-breaking

Callers that don't provide `tools` get identical behavior (full `claude_code` preset). `disableBuiltInTools: true` still works but is now a less flexible alias for `tools: []`.

## Test plan

- [x] All existing tests pass (`npm test` — 131 tests, 126 passed, 5 skipped)
- [x] New tests cover: default preset, explicit array, empty array, precedence over `disableBuiltInTools`
- [x] Verified `tools` type against [official SDK docs](https://platform.claude.com/docs/en/agent-sdk/typescript)

### DIFF
--- a/src/tests/acp-agent.test.ts
+++ b/src/tests/acp-agent.test.ts
@@ -1447,6 +1447,7 @@ describe("stop reason propagation", () => {
     agent.sessions["test-session"] = {
       query: messageGenerator() as any,
       input,
+      cwd: "/tmp/test",
       cancelled: false,
       permissionMode: "default",
       settingsManager: {} as any,


hancengiz:
@cla-bot check

PR #395, opened 2026-03-09
Move `clientCapability.auth._meta.gateway` to `clientCapability._meta.gateway`.

AlexandrSuhinin:
TS SDK filters out unknown fields. Field `auth` isn't passed to the `initialize` method.


benbrandt:
Solved by updating the TS SDK

PR #396, opened 2026-03-09
process SDK messages independently of prompt

SteffenDE:
This is a proof of concept of separating the SDK message loop from the prompt promise.

Relates to https://github.com/zed-industries/claude-agent-acp/pull/386.
Relates to https://github.com/zed-industries/claude-agent-acp/issues/336.


SteffenDE:
Note: I'll need to ensure /compact and /context still work, similar to #400.

benbrandt:
handling in #780

PR #403, opened 2026-03-09
fix: resolve model aliases in setSessionConfigOption (#401)

### LINKED ISSUE #401: Claude ACP session-level 'set model opus' fails with Internal error on 0.21.0
## Summary

`@zed-industries/claude-agent-acp` on the latest published release (`0.21.0`) still fails when switching an existing ACP session to **Opus** via session-level model config, even though:

- direct Claude CLI with Opus works
- Claude ACP with default/Sonnet works
- raw `claude-agent-acp` started with `ANTHROPIC_MODEL=opus` works

The failure only appears on the **session-level `set model opus` / `query.setModel("opus")` path**.

## Environment

- Package: `@zed-industries/claude-agent-acp`
- Installed version: `0.21.0`
- npm latest: `0.21.0`
- Host: macOS / OpenClaw ACPX bridge
- Claude auth: active / Claude Max

### Direct Claude CLI

```bash
claude --print --model sonnet 'Reply with exactly: DIRECT_CLI_SONNET_OK'
claude --print --model opus 'Reply with exactly: DIRECT_CLI_OPUS_OK'
claude --print --model claude-opus-4-6 'Reply with exactly: DIRECT_CLI_OPUS_FULL_OK'
```

All three succeed.

### Raw adapter pinned at process start

```bash
env ANTHROPIC_MODEL=opus npx -y @zed-industries/claude-agent-acp
```

Using the adapter this way works for Opus. I also verified this through ACPX with:

```bash
acpx --approve-all --non-interactive-permissions fail \
  --cwd /Users/.../workspace \
  --agent "env ANTHROPIC_MODEL=opus npx -y @zed-industries/claude-agent-acp" \
  exec 'Reply with exactly: RAW_ADAPTER_OPUS_OK'
```

Result:

```text
RAW_ADAPTER_OPUS_OK
```

### Session-level Sonnet switch

This works:

```bash
ACP=.../acpx
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude sessions ensure --name sonnet-retest >/dev/null
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude -s sonnet-retest set model sonnet >/dev/null
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude -s sonnet-retest exec 'Reply with exactly: ACP_CLAUDE_SONNET_RETEST_OK'
```

Result:

```text
ACP_CLAUDE_SONNET_RETEST_OK
```

## What fails

This fails consistently:

```bash
ACP=.../acpx
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude sessions ensure --name opus-retest >/dev/null
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude -s opus-retest set model opus >/dev/null
$ACP --approve-all --non-interactive-permissions fail --cwd /Users/.../workspace \
  claude -s opus-retest exec 'Reply with exactly: ACP_CLAUDE_OPUS_RETEST_OK'
```

Observed output:

```text
[acpx] created session opus-retest (...)
[acpx] agent: claude
[acpx] cwd: /Users/.../workspace
Internal error
```

This is reproducible across multiple fresh session names.

## Source inspection / why this seems surprising

I checked the installed `0.21.0` package and can confirm the code from the model-selection and better-error PRs is present locally:

### PR #340 (`Respect user settings for permission mode and model selection`)
Local code includes:
- tokenized model matching for aliases like `opus[1m]`
- `resolveModelPreference(...)`
- `getAvailableModels(...)`
- `currentModelId` handling

Relevant local lines in `dist/acp-agent.js`:
- `#L1053-L1115` (tokenized/alias model matching)
- `#L1117-L1134` (`getAvailableModels`, including `await query.setModel(currentModel.value)`)

### PR #366 (`Surface better error message when Claude Code process exits unexpectedly`)
Local code includes the improved process death checks:
- `#L506-L514`

So this does not appear to be “old package missing #340/#366”.

## Suspected failing line / path

The relevant path appears to be:

```js
if (params.configId === "model") {
  await this.sessions[params.sessionId].query.setModel(params.value);
}
```

in `dist/acp-agent.js` (around `#L560-L589`).

This suggests the break is specifically on:

- `query.setModel("opus")`

for a live session/query.

## Why I think this is a session-level model-switch bug, not a general Opus problem

Because all of these are true simultaneously:

- direct CLI Opus works
- raw adapter started with `ANTHROPIC_MODEL=opus` works
- session-level switch to Sonnet works
- session-level switch to Opus fails

That makes it look like a **live reconfiguration path** problem, not an availability/auth problem.

## Expected behavior

One of these should happen:

1. `set model opus` succeeds like Sonnet does, or
2. the adapter returns a clear, specific error explaining why Opus cannot be switched live, instead of generic `Internal error`.

## Actual behavior

- `set model opus` appears accepted
- next execution fails with only `Internal error`

## Possible fix directions

Two possibilities seem likely:

1. `query.setModel("opus")` is not safe for a live session/query and needs a reinit path for Opus-like targets
2. alias/canonicalization still resolves incorrectly for Opus in this runtime-switch path, even though startup-pinned Opus works

## If helpful

I can provide a tighter repro script and exact local file:line references, but the commands above should already reproduce it on the latest published package.

mudrii:
## Summary
Fixes #401 by enabling model alias/fuzzy resolution in `setSessionConfigOption` when updating the `model` config option.

## Root cause
`setSessionConfigOption` validated model values using an exact string match against config option values (model IDs like `claude-opus-4-5`):

```ts
const validValue = allValues.find((o) => o.value === params.value);
```

Clients often send human-friendly model preferences (`opus`, `sonnet`, `Claude Sonnet`), which failed this exact check and produced:

- `Invalid value for config option model: <value>`

The file already had `resolveModelPreference(...)`, which supports exact/display/substring/tokenized matching, but that logic was only used in model initialization paths and not in `setSessionConfigOption`.

### 1) Fuzzy model resolution in `setSessionConfigOption`
- Kept the existing exact-match validation path.
- When `configId === "model"` and exact match fails, map config option entries to `ModelInfo[]` (`value` + `displayName`) and call `resolveModelPreference(...)`.
- If resolution succeeds, use the resolved canonical model ID (`resolvedValue`) for:
  - `query.setModel(...)`
  - `session.configOptions` current value update
  - response payload (`configOptions`)

### 2) Notification behavior unchanged by design
- Intentionally **did not** add `updateConfigOption(...)` in `setSessionConfigOption`.
- This preserves existing behavior/tests that verify no `config_option_update` notification is sent for this RPC (response already returns full `configOptions`).

## Tests
Added focused coverage in `src/tests/session-config-options.test.ts` under `describe("setSessionConfigOption")`:
- resolves model alias `opus` -> `claude-opus-4-5`
- resolves model alias `sonnet` -> `claude-sonnet-4-5`
- resolves display name `Claude Sonnet` -> `claude-sonnet-4-5`
- still accepts exact model ID
- still throws for invalid model value (`gpt-4`)

### Before
- `setSessionConfigOption({ configId: "model", value: "opus" })` -> error

### After
- same call resolves to canonical ID and succeeds:
  - `setModel("claude-opus-4-5")`
  - returned `configOptions.model.currentValue === "claude-opus-4-5"`

## Verification
- `npm run build` ✅
- `npm test` ✅ (all tests passing)

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -923,27 +923,47 @@ export class ClaudeAcpAgent implements Agent {
       "options" in option && Array.isArray(option.options)
         ? option.options.flatMap((o) => ("options" in o ? o.options : [o]))
         : [];
-    const validValue = allValues.find((o) => o.value === params.value);
+    let validValue = allValues.find((o) => o.value === params.value);
+
+    // For model options, fall back to resolveModelPreference when the exact
+    // value doesn't match.  This lets callers use human-friendly aliases like
+    // "opus" or "sonnet" instead of full model IDs like "claude-opus-4-6".
+    if (!validValue && params.configId === "model") {
+      const modelInfos: ModelInfo[] = allValues.map((o) => ({
+        value: o.value,
+        displayName: o.name,
+        description: o.description ?? "",
+      }));
+      const resolved = resolveModelPreference(modelInfos, params.value);
+      if (resolved) {
+        validValue = allValues.find((o) => o.value === resolved.value);
+      }
+    }
+
     if (!validValue) {
       throw new Error(`Invalid value for config option ${params.configId}: ${params.value}`);
     }
 
+    // Use the canonical option value so downstream code always receives the
+    // model ID rather than the caller-supplied alias.
+    const resolvedValue = validValue.value;
+
     if (params.configId === "mode") {
-      await this.applySessionMode(params.sessionId, params.value);
+      await this.applySessionMode(params.sessionId, resolvedValue);
       await this.client.sessionUpdate({
         sessionId: params.sessionId,
         update: {
           sessionUpdate: "current_mode_update",
-          currentModeId: params.value,
+          currentModeId: resolvedValue,
         },
       });
     } else if (params.configId === "model") {
-      await this.sessions[params.sessionId].query.setModel(params.value);
+      await this.sessions[params.sessionId].query.setModel(resolvedValue);
     }
 
     session.configOptions = session.configOptions.map((o) =>
       o.id === params.configId && typeof o.currentValue === "string"
-        ? { ...o, currentValue: params.value }
+        ? { ...o, currentValue: resolvedValue }
         : o,
     );
 
--- a/src/tests/session-config-options.test.ts
+++ b/src/tests/session-config-options.test.ts
@@ -224,6 +224,61 @@ describe("session config options", () => {
       expect(configUpdate).toBeUndefined();
     });
 
+    it("resolves model alias 'opus' to full model ID", async () => {
+      const response = await agent.setSessionConfigOption({
+        sessionId: SESSION_ID,
+        configId: "model",
+        value: "opus",
+      });
+
+      expect(setModelSpy).toHaveBeenCalledWith("claude-opus-4-5");
+
+      const modelOption = response.configOptions.find((o) => o.id === "model");
+      expect(modelOption?.currentValue).toBe("claude-opus-4-5");
+    });
+
+    it("resolves model alias 'sonnet' to full model ID", async () => {
+      await agent.setSessionConfigOption({
+        sessionId: SESSION_ID,
+        configId: "model",
+        value: "sonnet",
+      });
+
+      expect(setModelSpy).toHaveBeenCalledWith("claude-sonnet-4-5");
+    });
+
+    it("resolves display name to model ID", async () => {
+      await agent.setSessionConfigOption({
+        sessionId: SESSION_ID,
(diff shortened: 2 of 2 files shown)


mudrii:
> We require contributors to sign our Contributor License Agreement, and we don't have @mudrii on file. You can sign our CLA at https://zed.dev/cla. Once you've signed, post a comment here that says '@cla-bot check'.

signed

PR #406, opened 2026-03-10
Support additional skill directories via additionalRoots meta parameter

egor-baranov:
Add support for `_meta.additionalRoots` in `claude-agent-acp`.

This change introduces a top-level `additionalRoots` field in session metadata and forwards it to Claude by merging it into `additionalDirectories` when creating a session.

Behavior:
- reads `_meta.additionalRoots` from the ACP session request
- treats it as a `string[]`
- merges it with any user-provided `claudeCode.options.additionalDirectories`
- passes the combined list through to Claude unchanged

This keeps `additionalRoots` as an ACP-level extension rather than a Claude-specific option, while mapping it onto Claude's existing `additionalDirectories` support.

### DIFF
--- a/src/tests/additional-roots.test.ts
+++ b/src/tests/additional-roots.test.ts
@@ -47,8 +47,9 @@ describe("additionalRoots", () => {
     } as unknown as AgentSideConnection);
   });
 
-  afterEach(async () =>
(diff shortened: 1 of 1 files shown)


egor-baranov:
@cla-bot check

PR #407, opened 2026-03-10
feat: add conversation forking support with message ID tracking

rohan-patra:
## Summary
- Adds message ID tracking to enable "edit previous message" UX via conversation forking
- Maps ACP user message IDs to preceding SDK assistant UUIDs, so `unstable_forkSession` can fork at a specific point in the conversation using `resumeSessionAt`
- Passes `messageId` through to ACP session notifications (text, image, thinking, tool_use chunks) so the client can correlate messages
- Returns `userMessageId` in prompt responses for client-side message tracking
- Uses user-provided `messageId` as the SDK message UUID (instead of random) when queuing messages during an active prompt

## Test plan
- [ ] Verify forking a session at a specific message replays conversation up to that point
- [ ] Verify editing a previous user message creates a forked session from the correct position
- [ ] Verify `messageId` appears on session update notifications
- [ ] Verify `userMessageId` is returned in all prompt response stop reasons
- [ ] Existing tests pass (messageIdMap added to test session fixtures)

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -117,6 +117,8 @@ type Session = {
   promptRunning: boolean;
   pendingMessages: Map<string, { resolve: (cancelled: boolean) => void; order: number }>;
   nextPendingOrder: number;
+  messageIdMap: Map<string, string>; // ACP user messageId -> preceding SDK assistant UUID
+  lastAssistantUuid?: string;
 };
 
 type BackgroundTerminal =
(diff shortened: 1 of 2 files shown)


SteffenDE:
Should probably wait for https://github.com/agentclientprotocol/agent-client-protocol/pull/629.

rohan-patra:
@SteffenDE yep, left it as a draft

PR #409, opened 2026-03-11
Add session/close support

SteffenDE:
~Waiting for https://github.com/agentclientprotocol/agent-client-protocol/pull/723.~


SteffenDE:
It fails because of the new [boolean session options](https://github.com/agentclientprotocol/agent-client-protocol/pull/576). Probably easy to fix, but I didn't look into it.

PR #410, opened 2026-03-12
Allow Bypass permissions mode after Exiting plan

Shivansh-25:
Adds a new `ExitPlanMode` choice so users can switch directly to `bypassPermissions` after accepting a plan, instead of only `acceptEdits` or `default`.

### Changes
- Updated the `ExitPlanMode` permission options in `src/acp-agent.ts` to conditionally include:
- `Yes, and bypass permissions` (`optionId: "bypassPermissions"`)
- Reused existing `ALLOW_BYPASS` guard so the option is only shown when bypass is supported.
- Updated selection handling so `bypassPermissions` is treated as a valid accepted outcome and applies mode updates like the existing paths.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -981,16 +981,25 @@ export class ClaudeAcpAgent implements Agent {
       }
 
       if (toolName === "ExitPlanMode") {
+        const options = [
+          {
+            kind: "allow_always",
+            name: "Yes, and auto-accept edits",
+            optionId: "acceptEdits",
+          },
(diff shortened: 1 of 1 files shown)


Shivansh-25:
@benbrandt This was an issue that I found in the zed-industries/zed repo and I am not sure of what the procedure is for getting the PR reviewed in this repo. 

I am sorry for tagging you out of the blue but you seem like an active contributor in this repo so you seemed like the right guy to ask. Thanks in advance :)

PR #363, opened 2026-03-01
fix: throw resourceNotFound when loadSession fails to resume

maskedband1t:
## Summary

- When `loadSession` calls `createSession({ resume: sessionId })` and the CLI subprocess can't find persisted session data (no `.jsonl` file), the subprocess exits immediately and `createSession` rejects with a generic `-32603` ("Query closed before response received")
- ACP clients like [acpx](https://github.com/openclaw/acpx) check for `-32001`/`-32002` (resource not found) to fall back to `session/new` — the generic `-32603` bypasses this, causing the turn to fail entirely
- This commonly happens when `session/new` creates an ACP session but the subprocess exits before writing session data (no prompt to process), then `session/prompt` tries `session/load` to resume a session that was never persisted

## Fix

Wrap the `createSession({ resume })` call in `loadSession()` with try/catch and re-throw as `RequestError.resourceNotFound()` (code `-32002`). This lets callers handle it with their existing fallback logic.

## Verification

Confirmed working in production with the [OpenClaw](https://github.com/openclaw/openclaw) gateway + acpx integration. Before the fix, all ACP tasks failed silently. After patching both this code path and acpx's `shouldFallbackToNewSession()` ([openclaw/acpx#30](https://github.com/openclaw/acpx/pull/30)), sessions fall back to `session/new` correctly.

Fixes #338
Closes #361
Closes #362

PR #413, opened 2026-03-12
fix: clear CLAUDECODE env to prevent nested-session detection

cyyij:
## Problem

When `claude-agent-acp` is spawned from within an existing Claude Code session (e.g. via `acpx` or other agent orchestrators), the parent process sets `CLAUDECODE=1`. This environment variable is inherited by the spawned subprocess, causing Claude Code to detect a "nested session" and exit immediately during initialization.

The result is a generic `-32603 Internal error` with `data.details: "Query closed before response received"` on every `session/new` call. This affects both `exec` (one-shot) and persistent session modes.

**Reproduction:**
```bash
# From inside a Claude Code session:
acpx claude exec 'hello'
# -> [error] RUNTIME: Internal error
```

## Root Cause

`CLAUDECODE=1` is set by Claude Code in its shell environment. When `claude-agent-acp` spawns a new Claude Code subprocess via the SDK's `query()`, it passes `...process.env` which includes `CLAUDECODE=1`. The child process sees this and refuses to start (nested-session protection).

## Fix

1. **Clear `CLAUDECODE` in subprocess env** — Set `CLAUDECODE: ""` in the env object passed to `query()`, so the spawned instance starts normally. This mirrors what other Claude Code wrappers do (e.g. cmux's wrapper script uses `unset CLAUDECODE`).

2. **Improve error handling for `session/new`** — The v0.20.0 fix for "Query closed before response received" (PR #363) only covered the `loadSession` path (`creationOpts.resume`). The same error on `newSession` was left unhandled, producing an opaque "Internal error". Now it returns a meaningful message suggesting to check Claude Code installation.

## Testing

Verified on macOS arm64 with:
- acpx 0.3.0
- Claude Code CLI 2.1.74
- Node v25.8.0

Before fix: `acpx claude exec` always fails with `Internal error` when run from within Claude Code.
After fix: works correctly, Claude responds normally.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -1227,6 +1227,7 @@ export class ClaudeAcpAgent implements Agent {
       ...userProvidedOptions,
       env: {
         ...process.env,
+        CLAUDECODE: "", // Prevent nested-session detection when spawned from within Claude Code
         ...userProvidedOptions?.env,
         ...createEnvForGateway(this.gatewayAuthMeta),
       },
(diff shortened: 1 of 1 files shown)


cyyij:
@cla-bot check

cyyij:
@cla-bot check

cyyij:
@cla-bot check

PR #423, opened 2026-03-15
Fix prompt hang when user message replay is absent

timvisher-dd:
> [!IMPORTANT]
> Claude thinks this isn't necessary if https://github.com/zed-industries/claude-agent-acp/pull/353 merges. I only became aware of the issue because my dev integration branch got reset to upstream/main and didn't have this replayed on top of it.

Commit 87a0886 (#400) introduced a `promptReplayed` flag to distinguish background task results from the current prompt's result. The flag starts `false` and only becomes `true` when the user message is replayed through the query stream with a matching UUID.

In practice, Claude Code either doesn't replay the user message or replays it with a different UUID. This causes `promptReplayed` to stay `false` for every prompt, so the result is always discarded as a "background task result". The `while(true)` loop then blocks forever waiting for a result that will never come — **hanging every prompt indefinitely**.

Wire logs from two independent sessions confirm:
1. `session/prompt` request is sent
2. Notifications flow normally (stream events, tool calls, agent messages)
3. stderr shows `"consuming background task result"` — the result is discarded
4. No response to the `session/prompt` request is ever sent
5. The client hangs forever (no timeout mechanism)

### Fix

Replace `promptReplayed` with `backgroundInitPending` — a flag that detects the actual fingerprint of a background task completion: a system `init` message immediately followed by a `result` with no intervening `stream_event` or `assistant` messages.

This works because:
- **Background tasks** produce `init → result` with nothing in between
- **Our prompt's processing** produces `init → stream_events → result` (Claude always streams its response before the result)
- The `stream_event` / `assistant` messages clear `backgroundInitPending`, so our result is accepted

This approach doesn't depend on user message replay at all, making it robust regardless of whether `replay-user-messages` is working correctly in the SDK.

### Tests added (7 new)

- [x] Result accepted when no user message replay arrives (the core regression)
- [x] Result accepted when `init` + stream events precede result without replay
- [x] `init → result` (no activity) correctly consumed as background task, followed by our result with stream events
- [x] Result accepted when stream events precede result without replay
- [x] Background task consumed when replay IS present (preserves original #400 intent)
- [x] Result accepted when replay has mismatched UUID
- [x] Existing background task test continues to pass
- [ ] Manual burn-in verification

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -488,7 +488,13 @@ export class ClaudeAcpAgent implements Agent {
     const promptUuid = randomUUID();
     userMessage.uuid = promptUuid;
 
-    let promptReplayed = false;
+    // backgroundInitPending tracks whether a system "init" message arrived
+    // without any subsequent activity (stream_event or assistant messages).
+    // A background task that completed after the previous prompt loop ended
+    // produces init → result with nothing in between. Our own prompt's
+    // processing always generates stream_event / assistant messages before
+    // its result, so the flag gets cleared before we see the result.
+    let backgroundInitPending = false;
 
     if (session.promptRunning) {
       session.input.push(userMessage);
@@ -499,9 +505,6 @@ export class ClaudeAcpAgent implements Agent {
       if (cancelled) {
         return { stopReason: "cancelled" };
       }
-      // The replay resolved the promise, mark in this loop too,
-      // so we don't treat the next result as a background task's result.
-      promptReplayed = true;
     } else {
       session.input.push(userMessage);
     }
@@ -524,6 +527,10 @@ export class ClaudeAcpAgent implements Agent {
           case "system":
             switch (message.subtype) {
               case "init":
+                // A new processing cycle started. If this is a
+                // background task, its result will follow immediately
+                // without intervening stream_event / assistant messages.
+                backgroundInitPending = true;
                 break;
               case "status": {
                 if (message.status === "compacting") {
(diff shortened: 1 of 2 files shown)


SteffenDE:
> In practice, Claude Code either doesn't replay the user message or replays it with a different UUID.

How did you determine this? User message replay works just fine for me with the exceptions of compaction and command outputs (where we manually set it to replayed).

> a system init message immediately followed by a result with no intervening stream_event or assistant messages.

That's not correct. I've seen Claude emit stream events and assistant messages for background tasks.

timvisher-dd:
> > In practice, Claude Code either doesn't replay the user message or replays it with a different UUID.
> 
> How did you determine this? User message replay works just fine for me with the exceptions of compaction and command outputs (where we manually set it to replayed).

I took wire logs from my sessions combined with observations in my `agent-shell` shell buffers and analyzed them with Claude. I can double check and provide the actual wire logs if that would be helpful. That said the tests that were added here fake the exact message structure that resulted in my broken state.

> > a system init message immediately followed by a result with no intervening stream_event or assistant messages.
> 
> That's not correct. I've seen Claude emit stream events and assistant messages for background tasks.

Definitely possible I misinterpreted the results. I'll try to provide further evidence if it becomes relevant. :)

benbrandt:
I think this should be fixed on main now?

SteffenDE:
I think so, yes!

SteffenDE:
@timvisher-dd please retry with 0.24.1 and report back :)

PR #431, opened 2026-03-16
fix session close method name

SteffenDE:
that's the nice part about unit tests, all green and still broken...

Relates to https://github.com/zed-industries/claude-agent-acp/pull/409.


SteffenDE:
Sorry @benbrandt, can you please include this before you do the next release?

PR #436, opened 2026-03-17
fix: suppress account assertion if gateway authentication is used

AlexandrSuhinin:


AlexandrSuhinin:
We need to suppress the auth_required error if we're going to use gateway authentication. Otherwise, we break migrating users who already have a subscription.

PR #437, opened 2026-03-17
fix context command

SteffenDE:
Closes https://github.com/zed-industries/claude-agent-acp/issues/435.

This is a draft. I need to do more testing.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -708,13 +708,17 @@ export class ClaudeAcpAgent implements Agent {
             // Slash commands like /compact can generate invalid output... doesn't match
             // their own docs: https://docs.anthropic.com/en/docs/claude-code/sdk/sdk-slash-commands#%2Fcompact-compact-conversation-history
             if (
-              typeof message.message.content === "string" &&
-              message.message.content.includes("<local-command-stdout>")
+              "model" in message.message &&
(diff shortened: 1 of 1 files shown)


josevalim:
In my opinion we should remove /context. We added this before we had session usage for Claude Code. So now this seems like unnecessary work to maintain it. (i.e. basically revert https://github.com/zed-industries/claude-agent-acp/pull/268).

benbrandt:
I'm also fine removing the context command

jgangemi:
isn't there still an issue w/ just returning formation from sub-agents or is that being handled someplace else?

SteffenDE:
Closing in favor of #432.

SteffenDE:
@jgangemi what do you mean with returning information from sub-agents? You can determine if an update is from a subagent or not by checking `_meta.claudeCode.parentToolUseId`.

jgangemi:
@SteffenDE i'm starting to think that maybe some of the issues i'm seeing are because of jetbrain's implementation of their plugin vs the adapter itself i already discovered that they use the the `.claude` directory by default instead of creating their own config dir which was clobbering my logins and potentially leading to the oauth issue i was having.

the fact that they keep releasing/changing things w/o any kind of announcements is highly frustrating too, side gripe.

Issue #441, opened 2026-03-19
Feature request: support effort and context window config options


hatawong:
## Summary

When using `claude-agent-acp` as an ACP adapter, there's no way for the ACP client to configure **effort** level or **context window size** through the ACP protocol.

### Effort
`buildConfigOptions()` only returns `mode` and `model` config options. Sending `session/set_config_option({ configId: "effort", value: "max" })` results in an error response because `effort` is not a recognized config option.

The Claude Code CLI supports `--effort low|medium|high|max`, and the Claude Agent SDK's `query()` likely supports effort configuration. But there's no ACP config option to pass it through.

### Context window
Context window size is determined by the model ID suffix `[1m]` (e.g., `claude-opus-4-6[1m]` → 1M context). The `resolveModelPreference()` function correctly parses `[1m]` hints via `tokenizeModelPreference()`, so passing `claude-opus-4-6[1m]` through `session/set_model` works.

However, this is fragile — it relies on the client knowing to include `[1m]` in the model ID. A dedicated context window config option (or at least documentation of this behavior) would be more robust.

## Expected behavior

- `buildConfigOptions()` should include an `effort` config option (values: `low`, `medium`, `high`, `max`)
- `setSessionConfigOption()` should handle `effort` and pass it to the Claude Agent SDK
- The `[1m]` model ID suffix behavior for context window should be documented

## Use case

We're building an AI agent orchestration platform ([aos gateway](https://github.com/hatawong/claude-recap)) that uses `claude-agent-acp` as an ACP adapter. Our behavior tree engine dispatches tasks with configurable model, effort, and context settings. Model works (via `unstable_setSessionModel`), but effort has no path through ACP.

## Workaround

For effort: We currently fall back to the CLI dispatch path (`claude -p --effort max`) which supports effort natively. ACP dispatch silently ignores effort failure (non-fatal).

For context: We pass the full model ID `claude-opus-4-6[1m]` which works via `resolveModelPreference()`.

## Environment

- `@zed-industries/claude-agent-acp`: latest (npm)
- `@agentclientprotocol/sdk`: 0.16.x
- Node.js 22+

benbrandt:
Hi yes we want this as well, but are still waiting for the SDK to support adjusting these on a running session. Once it does, we will add them!

PR #447, opened 2026-03-22
feat: skip setModel() when model is provided via query Options

hatawong:
## Summary

When a client passes `model` through `_meta.claudeCode.options` in `session/new`, the model is already set at CLI startup time via the `query()` Options parameter. The subsequent `setModel()` call in `getAvailableModels()` is unnecessary and causes two issues:

1. **`/model` command leaks into the ACP event stream** — `setModel()` triggers a `/model` CLI command whose output appears as `user_message_chunk` events, polluting the client's UI
2. **Extra round-trip** — adds a synchronous `setModel()` call to every session creation

## Changes

- Add `optionsModel` parameter to `getAvailableModels()` 
- When `optionsModel` is present, resolve it against available models for `currentModelId` reporting but **skip the `setModel()` call**
- Pass `options.model` (which comes from `_meta.claudeCode.options` spread) at the call site

## Effort support

The same `_meta.claudeCode.options` path also supports `effort` — it gets spread into SDK `Options` and takes effect at CLI startup via `query({ options: { effort } })`. No adapter changes needed for effort support; this PR unblocks it by establishing the pattern of passing config through `_meta` instead of post-hoc commands.

This addresses #441 — effort can now be passed through `_meta.claudeCode.options.effort` without waiting for a runtime `setEffort()` API.

## Backward compatibility

Clients that don't pass `_meta.claudeCode.options.model` get the existing behavior — `setModel()` is called as before. Only clients that explicitly provide a model via `_meta` benefit from the optimization.

## Test plan

- [ ] Existing tests pass
- [ ] Client sends `session/new` with `_meta: { claudeCode: { options: { model: "claude-opus-4-6" } } }` → no `/model` command in event stream
- [ ] Client sends `session/new` without `_meta` → existing `setModel()` behavior unchanged
- [ ] Client sends `_meta.claudeCode.options.effort` → effort takes effect at CLI startup

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -1425,7 +1425,7 @@ export class ClaudeAcpAgent implements Agent {
       );
     }
 
-    const models = await getAvailableModels(q, initializationResult.models, settingsManager);
+    const models = await getAvailableModels(q, initializationResult.models, settingsManager, options.model);
 
     const availableModes = [
       {
@@ -1619,6 +1619,7 @@ async function getAvailableModels(
   query: Query,
   models: ModelInfo[],
   settingsManager: SettingsManager,
+  optionsModel?: string,
 ): Promise<SessionModelState> {
   const settings = settingsManager.getSettings();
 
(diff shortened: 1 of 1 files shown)


hatawong:
@cla-bot check

benbrandt:
I believe this is already handled, thanks though!

PR #449, opened 2026-03-22
fix: only load current workspace's sessions instead of all sessions

0x5ea000000:
### Summary
When Zed (or another ACP client) opens the sessions list, it doesn't send a `cwd` parameter — the agent is spawned per-workspace, but the workspace path isn't passed at startup. Unlike Zed Agent, Claude Agent can only load sessions that were spawned under current workspace. We will encounter an error when opening sessions spawned from another workspace. I find this behavior unwelcome and annoying.

image

But since the `cwd` is passed from Zed side, there is no way the listSession() knows where current workspace is, I suggest a workaround. When a workspace is open, it opens a new agent session that includes the `cwd`, so we can cache the `cwd` to use later. 
### Changes
This PR caches the `cwd` from the first `newSession` call and uses it as a fallback in `listSessions`, so subsequent session list requests are scoped to the current workspace rather than showing all sessions across every project.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -275,6 +275,7 @@ export class ClaudeAcpAgent implements Agent {
   clientCapabilities?: ClientCapabilities;
   logger: Logger;
   gatewayAuthMeta?: GatewayAuthMeta;
+  workspaceCwd?: string;
 
   constructor(client: AgentSideConnection, logger?: Logger) {
     this.sessions = {};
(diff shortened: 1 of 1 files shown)


benbrandt:
This is a client problem of load cwd not matching original cwd. Will be fixed in zed shortly

0x5ea000000:
> This is a client problem of load cwd not matching original cwd. Will be fixed in zed shortly

Nice, thanks, really looking forward to the fix.

PR #455, opened 2026-03-23
fix: correct null check for gatewayAuthMeta in subscription validation

tbounsiar:
## Summary

- `gatewayAuthMeta` is declared as an optional property (`undefined` when unset), but the subscription validation compared it against `null`
- This made the condition always false, so `--hide-claude-auth` never blocked claude.ai subscription users when no gateway was configured
- Changed `=== null` to `!this.gatewayAuthMeta` to match the actual property semantics

## Test plan

- [x] `npm run lint` passes
- [x] All 6 authorization tests pass


tbounsiar:
@cla-bot check

tbounsiar:
@cla-bot check

PR #456, opened 2026-03-23
fix: include both stdout and stderr in Bash tool output

tbounsiar:
## Summary

- When a Bash tool execution produced both stdout and stderr, only stdout was shown due to the `||` operator short-circuiting
- Changed to concatenate both streams with `[stdout, stderr].filter(Boolean).join("\n")` so error output is no longer silently dropped

## Test plan

- [x] `npm run lint` passes
- [x] All 34 tools tests pass


tbounsiar:
@cla-bot check

tbounsiar:
@cla-bot check

PR #457, opened 2026-03-23
fix: always send valid used value in usage_update notification

### LINKED ISSUE #375: session/update usage_update returns Invalid params when used=null (via acpx 0.1.15)
### Summary
When using `acpx` with the `claude` agent, a successful run still prints a protocol warning on stderr:

- notification: `session/update`
- update type: `usage_update`
- payload contains `used: null`
- receiver responds: `code: -32602, message: Invalid params`

The prompt result is correct and exit code is `0`, but we expect no protocol warnings in normal operation.

### Environment
- OS: macOS (Darwin 25.3.0)
- Node: v25.6.1
- acpx: `0.1.15`
- claude-agent-acp: latest from npm at test time (`npx -y @zed-industries/claude-agent-acp`)

### Reproduction
```bash
cd /usr/local/lib/node_modules/openclaw/extensions/acpx
./node_modules/.bin/acpx --cwd /Users/mac/.openclaw/workspace --format quiet --timeout 60 claude exec "Return exactly ACP_SMOKE_OK"
```

### Actual output
```text
ACP_SMOKE_OK
Error handling notification {
  jsonrpc: '2.0',
  method: 'session/update',
  params: {
    sessionId: '... ',
    update: {
      sessionUpdate: 'usage_update',
      used: null,
      size: 200000,
      cost: [Object]
    }
  }
} {
  code: -32602,
  message: 'Invalid params',
  data: {
    ...
    used: { _errors: [Array] }
  }
}
```

### Notes
- Exit code is still `0`.
- Same smoke check with `codex` does **not** emit this warning.
- This looks like a schema mismatch around nullable `used` in `usage_update`.

### Expected
Either:
1. `used` is always sent as a valid object, or
2. `used: null` is accepted by the receiving schema for `usage_update`.

No protocol warning should appear for successful normal runs.

tbounsiar:
Fixes #375

## Summary

- When no assistant message with usage was seen before a `result` message (e.g. `output_tokens=0` or compact-only turns), `lastAssistantTotalUsage` remained `null`
- Although guarded by `if (lastAssistantTotalUsage !== null)`, the usage_update was simply skipped — some clients still received `used: null` causing `Invalid params` errors
- Now always sends the `usage_update` with `used: 0` as fallback when no usage data is available

## Test plan

- [x] `npm run lint` passes
- [x] All tests pass (2 pre-existing Windows path separator failures unrelated)

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -570,21 +570,20 @@ export class ClaudeAcpAgent implements Agent {
             const contextWindowSize =
               contextWindows.length > 0 ? Math.min(...contextWindows) : 200000;
 
-            // Send usage_update notification
-            if (lastAssistantTotalUsage !== null) {
-              await this.client.sessionUpdate({
-                sessionId: params.sessionId,
-                update: {
-                  sessionUpdate: "usage_update",
-                  used: lastAssistantTotalUsage,
-                  size: contextWindowSize,
-                  cost: {
-                    amount: message.total_cost_usd,
-                    currency: "USD",
-                  },
+            // Send usage_update notification — always send with a valid
+            // number so clients don't receive `used: null` (see #375).
+            await this.client.sessionUpdate({
+              sessionId: params.sessionId,
+              update: {
+                sessionUpdate: "usage_update",
+                used: lastAssistantTotalUsage ?? 0,
+                size: contextWindowSize,
+                cost: {
+                  amount: message.total_cost_usd,
+                  currency: "USD",
                 },
-              });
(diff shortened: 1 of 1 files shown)


SteffenDE:
As mentioned in the linked issue, this seems to be an upstream issue. We only send a usage update when we actually have data, which I'd say is the correct approach.

tbounsiar:
Closing — agreed this is an upstream issue, not something to fix on the agent side. Thanks for the clarification!

PR #464, opened 2026-03-25
feat: Support effort levels

benbrandt:


benbrandt:
It's wired up... but I am not convinced it is working on my plan... Feels like we need some other check to know if the user can set this or not...

kayakyakr:
This seems to be unnecessary now as 0.29.2, at least, comes with functional effor level selection.

martenlienen:
Can you describe how? According to the changelog, 0.29.2 only has a bugfix that seems unrelated.

kayakyakr:
> Can you describe how? According to the changelog, 0.29.2 only has a bugfix that seems unrelated.

My bad. It looks like it was something that @rohan-patra added in his 0.27.0 update. I'm running my own customized fork and I must have gotten confused with which codebase brought in which code.

It might be worth taking a look as his implementation, seeing how it differs, the effort_level selector code in his branch is clean:
https://github.com/rohan-patra/claude-agent-acp/commit/151a99140cbbff00f7841c3d528dfe0f0275aa53

lixiaoyan:
Using the locally built version, I noticed that the effort level is not displayed initially; it only appears after the model is manually switched.

https://github.com/user-attachments/assets/759e37eb-0727-4e46-bb9b-09ea6fae6417

slonopotamus:
Effort level always gets reset to `xhigh` when I open new chat. Using Claude Agent 0.31.0 with Jetbrains IDEA.

PR #498, opened 2026-04-01
fix: allow opt-out of session_state_changed via env var

### LINKED ISSUE #497: prompt() hangs when Claude Code binary doesn't support CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS
## Bug Description

Commit 7f1c6c20 ("Use idle session state as end of turn (#463)") and its follow-up 23b3073 ("Cleanup based on new idle state #463 (#480)") moved the prompt loop's exit from the `result` message handler to `session_state_changed(idle)`. This relies on the Claude Code binary supporting the `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS` env var.

When the Claude Code binary **does not** support this feature (e.g., older versions), `session_state_changed(idle)` is never emitted. The prompt loop calls `session.query.next()` after processing the `result` message and **blocks forever**, because the process is still alive but sends no further messages.

This causes the JSON-RPC response for `session/prompt` to never be sent, and the client's `conn.prompt()` future never resolves — **deadlocking the client**.

## Steps to Reproduce

1. Use `claude-agent-acp` v0.23.0+ (which includes commit 7f1c6c20)
2. Connect with a Claude Code binary that does not support `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS` (e.g., Claude Code 2.1.77 installed by npm, set env CLAUDE_CODE_EXECUTABLE to point to it)
3. Send a `session/prompt` request
4. The agent processes the turn, sends streaming notifications, receives `result(success)` — but never returns the `PromptResponse`
5. Client hangs indefinitely

## Root Cause

In the prompt loop (`acp-agent.ts:prompt()`), all `result` subtypes (`success`, `error_during_execution`, `error_max_*`) now set `stopReason` and `break` instead of `return`. The only return path is via `session_state_changed(idle)`, which requires the env var to be supported by the binary:

```typescript
// Result handler — no longer returns
case "success": {
    // ... error checks ...
    break;  // was: return { stopReason: "end_turn", usage }
}

// Only exit path
case "session_state_changed": {
    if (message.state === "idle") {
        return { stopReason, usage: sessionUsage(session) };  // never reached
    }
}
```

After `break`, the loop calls `await session.query.next()` which blocks forever since the process sends no more messages.

## Expected Behavior

The prompt loop should return a `PromptResponse` even when `session_state_changed` is not available.

## Environment

- claude-agent-acp: v0.23.0+ (commits 7f1c6c20, 23b3073)
- Claude Code binary: any version that doesn't support `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS`

vincentye38:
## Summary

- Fixes #497
- When `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS=0` is set in the environment, the prompt loop returns the `PromptResponse` directly from the `result` message handler instead of waiting for `session_state_changed(idle)` which will never arrive on older Claude Code binaries
- The env var defaults to `"1"` (enabled) to preserve existing behavior
- Respects user's env var instead of hardcoding `"1"`, allowing deterministic opt-out

## Changes

1. **`Session.useSessionStateEvents`** — new boolean flag on the session, derived from the env var at session creation
2. **`createSession` env override** — `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS` now respects `process.env` instead of always hardcoding `"1"`
3. **Result handler fallback** — after the inner result subtype switch, returns immediately when `useSessionStateEvents` is `false`

## Test plan

- [x] With `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS` unset (default): behavior unchanged, waits for `session_state_changed(idle)`
- [x] With `CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS=0`: prompt returns immediately after `result(success)`, no hang
- [x] With older Claude Code binary + env var set to `0`: client receives `PromptResponse` and does not deadlock

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -123,6 +123,7 @@ type Session = {
   pendingMessages: Map<string, { resolve: (cancelled: boolean) => void; order: number }>;
   nextPendingOrder: number;
   abortController: AbortController;
+  useSessionStateEvents: boolean;
 };
 
 type BackgroundTerminal =
@@ -681,6 +682,12 @@ export class ClaudeAcpAgent implements Agent {
               default:
                 unreachable(message, this.logger);
                 break;
+            }
+            // When session state events are disabled (CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS=0),
+            // session_state_changed(idle) will never arrive. Return immediately
+            // using the stopReason set by the result subtype above.
+            if (!session.useSessionStateEvents) {
+              return { stopReason, usage: sessionUsage(session) };
             }
             break;
           }
@@ -1364,8 +1371,11 @@ export class ClaudeAcpAgent implements Agent {
         ...process.env,
         ...userProvidedOptions?.env,
         ...createEnvForGateway(this.gatewayAuthMeta),
-        // Opt-in to session state events like when the agent is idle
-        CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS: "1",
+        // Allow users to opt out of session state events by setting
+        // CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS=0 in their environment,
+        // e.g. when the Claude Code binary does not support this feature.
+        CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS:
+          process.env.CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS ?? "1",
       },
       // Override certain fields that must be controlled by ACP
       cwd: params.cwd,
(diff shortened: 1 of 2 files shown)


josevalim:
Do we want to support different versions of Claude Code or should we instead document that you need a minimum version? Especially because this seems to fix the problem only by explicitly setting an env var, which I assume has low discovery?

vincentye38:
CLAUDE_CODE_EMIT_SESSION_STATE_EVENTS Is introduced in v2.1.83. The default is set to 0. IMO, We should maintain this acp agent downward compatible with 0 while supporting 1 until Claude Code SDK changes default setting to 1.

vincentye38:
@benbrandt Could you take a look at this PR. Thank you!

benbrandt:
Hey @vincentye38 I don't think we want to do this. We moved to these idle events to deal with broken behavior from the drop events. 

We can only support custom claude executables so much, usually it is because of internal requirements, but hopefully you are able to update to a newer version (at this point you are at least 8 releases behind if you aren't on at least 2.1.83). Adding this PR would bring back broken behavior of the sessions seeming to get stuck

PR #506, opened 2026-04-05
fix: emit usage updates from Claude stream events

Mng-dev-ai:
## Summary
- emit ACP `usage_update` notifications from top-level Claude `stream_event` message usage
- keep tracking a usage snapshot across `message_start` and `message_delta` so partial deltas still produce a complete total
- preserve the existing `result`-path update as the final exact/cost-bearing usage event

## Why
Claude SDK stream events already carry incremental usage on `message_delta`, but `claude-agent-acp` currently ignores that path and only emits `usage_update` from the final `result`. This makes usage appear stale until the turn ends in ACP clients that expect live context usage updates.

## Verification
- built the repo with `npm run build`
- ran a direct ACP stdio client against the patched local build
- observed `usage_update` notifications emitted before the final result-path usage update

Patched local smoke-test output included this sequence:
- `usage_update used=20973 size=200000`
- `agent_message_chunk text=TEST_OK`
- `usage_update used=20989 size=200000`
- `usage_update used=20989 size=1000000 cost=0.1314875`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -115,12 +115,14 @@ type UsageSnapshot = {
   cache_creation_input_tokens: number;
 };
 
-const ZERO_USAGE: UsageSnapshot = {
+const ZERO_USAGE: Readonly<UsageSnapshot> = Object.freeze({
   input_tokens: 0,
   output_tokens: 0,
   cache_read_input_tokens: 0,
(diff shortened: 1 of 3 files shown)


Mng-dev-ai:
@benbrandt sorry for bothering you but just want to know if you're accepting PRs or not

PR #508, opened 2026-04-05
Forward dropped SDK events to ACP clients

m13v:
## Summary

The claude-agent-acp runtime currently drops several Claude Agent SDK messages on the floor: `api_retry`, `tool_progress`, `tool_use_summary`, and `rate_limit_event`. ACP clients can't see that the SDK is retrying a 529, that a long-running tool is making progress, that the model summarized prior tool uses, or that a Claude.ai web-session rate limit fired.

This PR forwards those four events to clients via typed `_claude/`-namespaced `extNotification` calls, so clients that care can surface them without touching the transport.

## What's forwarded

All four events are forwarded via `this.client.extNotification("_claude/<name>", { sessionId, ...fields })`:

1. **`_claude/api-retry`** — fired when the SDK retries a transient API error (overload, rate limit response, etc.)
   - `httpStatus`, `errorType`, `attempt`, `maxRetries`, `retryDelayMs`

2. **`_claude/tool-progress`** — periodic heartbeat from a long-running tool
   - `toolUseId`, `toolName`, `elapsedTimeSeconds`

3. **`_claude/tool-use-summary`** — model-generated summary of a group of prior tool uses
   - `summary`, `precedingToolUseIds`

4. **`_claude/rate-limit`** — Claude.ai web-session rate limit state change
   - `status`, `resetsAt`, `rateLimitType`, `utilization`, `overageStatus`, `overageDisabledReason`, `isUsingOverage`, `surpassedThreshold`

## Why `extNotification`

Per reviewer feedback, these aren't first-class ACP session updates (the spec doesn't define them), so forwarding them as `sessionUpdate` entries would pollute the standard update stream. `extNotification` with a `_claude/` prefix is the documented escape hatch for vendor-specific events, keeps the core ACP surface clean, and lets clients opt in by listening on the namespaced method.

## Other changes

- **Fall-through bug fix**: the previous switch had `hook_progress`, `hook_response`, `files_persisted`, `task_started`, `task_notification`, `task_progress`, and `elicitation_complete` falling through into the newly-added `api_retry` handler. Each now has its own explicit `break`.
- **Dropped `sessionMeta()` / `_meta` additions** from earlier revisions — no longer needed once events are forwarded directly to the client as they happen.
- Logs `api_retry` at error level (same as before) so the CLI still shows retries in the terminal.

## Field naming

Field names match the fazm ACP client's existing shape (`toolUseId`, `httpStatus`, `retryDelayMs`, `precedingToolUseIds`, etc.) so a client already consuming these from a monkey-patched bridge can swap over without changing its handler code.

## Testing

- `npm run build` — clean
- `npm test` — all existing tests pass; mock Session objects updated to drop the removed `totalCostUsd` / `lastTerminalReason` fields

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -116,10 +116,6 @@ type Session = {
   cwd: string;
   settingsManager: SettingsManager;
   accumulatedUsage: AccumulatedUsage;
-  /** Cumulative cost in USD for this session (from SDK result messages) */
-  totalCostUsd: number;
-  /** Terminal reason from the last SDK result (e.g. "blocking_limit", "completed") */
-  lastTerminalReason: string | null;
   modes: SessionModeState;
   models: SessionModelState;
   configOptions: SessionConfigOption[];
@@ -490,9 +486,6 @@ export class ClaudeAcpAgent implements Agent {
       cachedReadTokens: 0,
       cachedWriteTokens: 0,
     };
-    session.totalCostUsd = 0;
-    session.lastTerminalReason = null;
-
     let lastAssistantTotalUsage: number | null = null;
     let lastAssistantModel: string | null = null;
     let lastContextWindowSize: number = 200000;
@@ -596,11 +589,7 @@ export class ClaudeAcpAgent implements Agent {
               }
               case "session_state_changed": {
                 if (message.state === "idle") {
-                  return {
-                    stopReason,
-                    usage: sessionUsage(session),
-                    _meta: sessionMeta(session),
-                  };
+                  return { stopReason, usage: sessionUsage(session) };
                 }
                 break;
               }
@@ -612,30 +601,23 @@ export class ClaudeAcpAgent implements Agent {
               case "task_notification":
               case "task_progress":
               case "elicitation_complete":
+                break;
               case "api_retry": {
-                // Forward API retry events so clients can observe transient failures,
-                // including the HTTP status code and typed error category.
+                // Forward API retry events via extNotification so clients can observe
+                // transient failures including HTTP status code and typed error category.
                 const retryMsg = message as any;
                 this.logger.error(
                   `API retry: attempt=${retryMsg.attempt}/${retryMsg.max_retries}, ` +
                   `httpStatus=${retryMsg.error_status}, error=${retryMsg.error}, ` +
                   `delay=${retryMsg.retry_delay_ms}ms`,
                 );
-                await this.client.sessionUpdate({
-                  sessionId: retryMsg.session_id,
-                  _meta: {
-                    apiRetry: {
-                      httpStatus: retryMsg.error_status ?? null,
-                      errorType: retryMsg.error ?? "unknown",
-                      attempt: retryMsg.attempt ?? 0,
-                      maxRetries: retryMsg.max_retries ?? 0,
-                      retryDelayMs: retryMsg.retry_delay_ms ?? 0,
-                    },
-                  },
-                  update: {
-                    sessionUpdate: "agent_message_chunk",
-                    content: { type: "text", text: "" },
-                  },
+                await this.client.extNotification("_claude/api-retry", {
+                  sessionId: params.sessionId,
+                  httpStatus: retryMsg.error_status ?? null,
+                  errorType: retryMsg.error ?? "unknown",
+                  attempt: retryMsg.attempt ?? 0,
+                  maxRetries: retryMsg.max_retries ?? 0,
+                  retryDelayMs: retryMsg.retry_delay_ms ?? 0,
                 });
                 break;
               }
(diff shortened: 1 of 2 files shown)


src/acp-agent.ts
```diff
@@ -616,6 +650,10 @@ export class ClaudeAcpAgent implements Agent {
             session.accumulatedUsage.outputTokens += message.usage.output_tokens;
             session.accumulatedUsage.cachedReadTokens += message.usage.cache_read_input_tokens;
             session.accumulatedUsage.cachedWriteTokens += message.usage.cache_creation_input_tokens;
+            session.totalCostUsd = message.total_cost_usd;
```
SteffenDE: We already include total_cost_usd in the `usage_update` message.

src/acp-agent.ts
```diff
@@ -835,11 +873,69 @@ export class ClaudeAcpAgent implements Agent {
...
+            if (toolCallId) {
+              await this.client.sessionUpdate({
+                sessionId: params.sessionId,
+                update: {
+                  sessionUpdate: "tool_call_update",
+                  toolCallId,
+                  status: "running",
+                  content: [{
+                    type: "text",
+                    text: `Tool running (${Math.round((message as any).elapsed_time_seconds ?? 0)}s)...`,
```
SteffenDE: I wouldn't do this. Content for a tool call should be based on the actual tool output. If you want progress, I'd recommend to send it as `extNotification("_claude/tool-progress", ...)` or similar.

Now, I'm thinking that instead of adding custom mappings for many more SDK messages, what if we'd allow you to pass a custom flag `includeRawSDKMessages` in `session/new` etc. and we'd then forward all SDK messages as custom session notifications, so you'd be able to handle them however you want in your client?

Then, in the future we might have "proper" ACP messages for errors / rate-limits and you could switch to those at some point.

src/acp-agent.ts
```diff
@@ -602,9 +612,33 @@ export class ClaudeAcpAgent implements Agent {
...
+                      errorType: retryMsg.error ?? "unknown",
+                      attempt: retryMsg.attempt ?? 0,
+                      maxRetries: retryMsg.max_retries ?? 0,
+                      retryDelayMs: retryMsg.retry_delay_ms ?? 0,
+                    },
+                  },
+                  update: {
+                    sessionUpdate: "agent_message_chunk",
+                    content: { type: "text", text: "" },
+                  },
```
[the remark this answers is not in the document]
m13v: Good call. Revised to use `extNotification("_claude/api-retry", ...)` and similarly `_claude/rate-limit`, `_claude/tool-progress`, `_claude/tool-use-summary` for all four dropped events.

src/acp-agent.ts
m13v: You're right, removed. The `sessionMeta()` helper and the `_meta` additions to return values are gone. `total_cost_usd` is already carried by `usage_update`.

src/acp-agent.ts
m13v: Agreed. Switched to `extNotification("_claude/tool-progress", ...)` — passes `toolCallId`, `toolName`, and `elapsedTimeSeconds` as notification params rather than injecting text into the tool output content.

src/acp-agent.ts
```diff
@@ -1569,6 +1667,18 @@ function sessionUsage(session: Session) {
...
 }
 
+/** Build _meta for PromptResponse with cost and terminal reason from SDK results */
+function sessionMeta(session: Session): Record<string, unknown> {
+  const meta: Record<string, unknown> = {};
+  if (session.totalCostUsd > 0) {
+    meta.costUsd = session.totalCostUsd;
+  }
+  if (session.lastTerminalReason) {
+    meta.terminalReason = session.lastTerminalReason;
```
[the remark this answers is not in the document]
m13v: Since we switched fully to `extNotification`, the `_meta` nesting question is moot for these events — the method name (`_claude/rate-limit` etc.) provides the namespace. The `_meta: { claudeCode: ... }` pattern is preserved for the existing `sessionUpdate` calls that already use it.

m13v:
On the `includeRawSDKMessages` suggestion (and your draft in #527) — that's a cleaner long-term shape than forwarding one event at a time, and I'd happily use it if it lands. The reason I kept this PR narrow to the four specific events is that they're the ones I've hit in practice where the current behaviour is strictly lossy: the SDK has information the client needs, the runtime sees it, and then drops it. Forwarding them as typed `_claude/` notifications means clients that already care about these cases can consume them without having to opt into the full raw message firehose or parse SDK internals.

If #527 moves forward, this PR's handlers should be easy to either drop or rebase on top of it (they're small and self-contained). Happy to do whatever's cleanest — delete these in favor of `includeRawSDKMessages`, keep them as typed convenience notifications alongside it, or close this in favor of #527 entirely. Let me know which direction you'd prefer.

benbrandt:
Let's see if #527 solves your issue and then we can revisit if not

PR #523, opened 2026-04-08
fix: Use TUI login for remote environments

benbrandt:
The auth subcommand goes through a flow that requires pasting in a code on remote environments, but does not allow for pasting that code in. Until they fix this in the CLI subcommand, we'll have to go to the old flow of launching the full CLI.


benbrandt:
cc @SteffenDE

SteffenDE:
Hmm interesting. Not sure if that helps in my case. I don't have a NO_BROWSER env var set.

SteffenDE:
Update: it's good, we just always set NO_BROWSER for now :)

benbrandt:
Hmm i wonder if we can do a similar check to what they are doing to decide which url to generate... then it would hopefully be more consistent...

PR #530, opened 2026-04-08
fix: exit process when ACP connection closes

Sway-Chan:
## Summary

- Return `AgentSideConnection` from `runAcp()` instead of discarding it
- Listen for `connection.closed` and call `process.exit(0)` on resolution
- Fixes orphan process accumulation in oneshot mode (e.g. via ACPX)

## Problem

When used in oneshot mode, `claude-agent-acp` does not exit after the client closes the stdin pipe. `process.stdin.resume()` keeps the Node.js event loop alive indefinitely. Over time, orphan processes accumulate on the host (~97MB RSS each), eventually exhausting system resources.

## Root cause

`runAcp()` creates an `AgentSideConnection` but discards the reference. The `Connection.#receive()` loop detects stdin EOF and fires `abortController.abort()`, resolving `connection.closed` — but no code listens for it to trigger `process.exit()`.

## Fix

1. **`acp-agent.ts`**: `runAcp()` now returns the `AgentSideConnection` instance
2. **`index.ts`**: Awaits `connection.closed` and calls `process.exit(0)`

## Impact

- **Oneshot sessions**: Process now exits cleanly when the client closes stdin (EOF)
- **Persistent sessions**: Unaffected — the client keeps stdin open, so `connection.closed` does not fire until explicit disconnect

## Test plan

- [x] Verified oneshot: process exits within 1s after stdin EOF
- [x] Verified persistent: process remains alive while stdin is open
- [x] Measured idle RSS: ~97MB → freed on exit
- [x] Tested with ACPX runtime (OpenClaw gateway) in production

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -2239,7 +2239,7 @@ export function runAcp() {
   const output = nodeToWebReadable(process.stdin);
 
   const stream = ndJsonStream(input, output);
-  new AgentSideConnection((client) => new ClaudeAcpAgent(client), stream);
+  return new AgentSideConnection((client) => new ClaudeAcpAgent(client), stream);
 }
 
 function commonPrefixLength(a: string, b: string) {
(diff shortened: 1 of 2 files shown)


benbrandt (approved):
thanks!

PR #527, opened 2026-04-08
allow clients to opt into receiving raw SDK messages

SteffenDE:
Relates to: https://github.com/agentclientprotocol/claude-agent-acp/pull/508#discussion_r3050540393

PR #532, opened 2026-04-09
Surface rate_limit_event quota metadata

suxxes:
Emit a quota _meta payload for rate_limit_event so ACP clients can track rate limits without parsing SDK-specific stream messages. This keeps the change isolated to ACP session updates.


josevalim:
Maybe related to #527.

benbrandt:
Closing in favor of #527

PR #534, opened 2026-04-11
Add CLAUDE_CODE_WRAPPER_MODE environment variable support for enterprises using Claude Code wrappers that handle authentication internally

jiaohanying:
Add CLAUDE_CODE_WRAPPER_MODE environment variable support for enterprises using Claude Code wrappers that handle authentication internally (e.g. enterprise SSO, API gateway). When enabled, all ACP-level auth prompts are skipped and the configured CLAUDE_CODE_EXECUTABLE is used directly.

CLAUDE_CODE_WRAPPER_MODE requires CLAUDE_CODE_EXECUTABLE to be set, and throws a descriptive error if misconfigured.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -240,6 +240,21 @@ function shouldHideClaudeAuth(): boolean {
   return process.argv.includes("--hide-claude-auth");
 }
 
+function shouldUseWrapperMode(): boolean {
+  if (process.env.CLAUDE_CODE_WRAPPER_MODE) {
+    if (!process.env.CLAUDE_CODE_EXECUTABLE) {
+      throw new Error(
+        "CLAUDE_CODE_WRAPPER_MODE requires CLAUDE_CODE_EXECUTABLE to be set. " +
+        "Please configure CLAUDE_CODE_EXECUTABLE to point to your Claude Code wrapper executable " +
+        "that handles authentication internally (e.g. enterprise SSO, API gateway). " +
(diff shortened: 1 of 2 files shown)


josevalim:
Perhaps it should be called `CLAUDE_CODE_CUSTOM_AUTH` if it is only changes the authentication? 🤔

jiaohanying:
`CLAUDE_CODE_WRAPPER_MODE` has been renamed to `CLAUDE_CODE_CUSTOM_AUTH`

src/acp-agent.ts
```diff
@@ -425,12 +440,15 @@ export class ClaudeAcpAgent implements Agent {
...
+        : [...terminalAuthMethods, ...(supportsGatewayAuth ? [gatewayAuthMethod] : [])],
     };
   }
 
   async newSession(params: NewSessionRequest): Promise<NewSessionResponse> {
     if (
+      !shouldUseWrapperMode() &&
       !this.gatewayAuthMeta &&
       fs.existsSync(path.resolve(os.homedir(), ".claude.json.backup")) &&
       !fs.existsSync(path.resolve(os.homedir(), ".claude.json"))
```
benbrandt, line 454: I wonder if the right thing to do here is just remove this check... It's kind of an optimistic one and we should get the auth required error once the prompt is submitted. I assume this would also solve your issue?

benbrandt:
I assume https://github.com/agentclientprotocol/claude-agent-acp/pull/544 will solve your issue. Let me know if not

PR #536, opened 2026-04-13
fix: guard tool info rendering when tool_use input is undefined

DASungta:
## Summary

- Guard `toolInfoFromToolUse()` against `undefined` input for Read, Glob, Grep, WebSearch, and TodoWrite cases
- Make `planEntries()` resilient to `undefined` input (returns `[]` instead of crashing)
- Guard TodoWrite plan updates in `streamEventToAcpNotifications()` with optional chaining on `chunk.input?.todos`

## Why this change

Claude `tool_use` blocks can temporarily carry `input: undefined` during streaming (before `input_json_delta` arrives), and historical `tool_use` messages with incomplete input may be replayed when a JetBrains IDEA session is resumed. Without these guards, ACP crashes with:

```
TypeError: Cannot read properties of undefined (reading 'limit')
TypeError: Cannot read properties of undefined (reading '-i')
TypeError: Cannot read properties of undefined (reading 'todos')
```

All three crashes share the same root cause: the ACP bridge assumes `tool_use.input` is always defined, but it can be `undefined` during streaming early chunks or session history replay.

## Test plan

- [x] Added regression tests for `toolInfoFromToolUse()` with `input: undefined` (Read, Grep, Glob, WebSearch, TodoWrite)
- [x] Added regression test for `planEntries(undefined)` — returns `[]`
- [x] Added regression test for `planEntries({})` — graceful handling
- [x] Added regression test for `toAcpNotifications()` with TodoWrite + undefined input — no crash, no spurious plan update
- [x] Added regression test verifying TodoWrite + valid input still emits plan update
- [x] `npm run build` passes with no TypeScript errors
- [x] All 167 tests pass (10 new)

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -2002,8 +2002,8 @@ export function toAcpNotifications(
         const alreadyCached = chunk.id in toolUseCache;
         toolUseCache[chunk.id] = chunk;
         if (chunk.name === "TodoWrite") {
-          // @ts-expect-error - sometimes input is empty object
-          if (Array.isArray(chunk.input.todos)) {
+          // @ts-expect-error - sometimes input is empty object or undefined
+          if (Array.isArray(chunk.input?.todos)) {
             update = {
               sessionUpdate: "plan",
               entries: planEntries(chunk.input as { todos: ClaudePlanEntry[] }),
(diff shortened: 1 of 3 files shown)


iwangjie:
I encountered the same problem and am waiting for it to be fixed.

DASungta:
299418797d8d0a96bfb2dccf8023d0a2

benbrandt (approved):
thanks!

PR #538, opened 2026-04-13
fix: Better remote check for auth methods

benbrandt:
The auth methods don't have interactive modes, which means they don't
work on remote machines when you need to paste a code in.

Our previous change might not have been enough, so added some heuristics
similar to what the SDK does for checking for remote.

It's not perfect because I think they are attempting to open the URL and
then falling back when it fails, but heopfully this is at least closer.


benbrandt:
cc @SteffenDE not sure if this still helps in your case...

SteffenDE:
Looks good to me. For Tidewave we always want to use the interactive terminal anyway, so it doesn't really make a difference for us (setting NO_BROWSER is fine!)

PR #568, opened 2026-04-19
feat: add Usage config dropdown for context and rate-limit visibility

kayakyakr:
## Summary

Adds a display-only session config option, `Usage`, that surfaces live context-window and rate-limit usage as a dropdown. The option's `currentValue` picks which metric is shown on the closed dropdown button:

- `Ctx: 45%` - context window
- `5h 62% 2:15` — utilization and/or time until reset as `H:MM`
- `Week: 18%` — Opus weekly; Sonnet appended in the tooltip
- `(hidden)` — empty label, collapses to just the chevron

Per-option `description` fields carry the verbose details (absolute token counts, reset times with timezone, approaching-limit status, weekly Opus vs Sonnet) — Zed renders these as tooltips.

## Why as a session config option

`SessionConfigOption` is the only spot in the current ACP spec that lets an agent push a persistent, user-visible indicator into the client UI without requiring a client-side change. `SessionUpdate::UsageUpdate` already covers context tokens, but:

- It's `UNSTABLE` and gated behind `AcpBetaFeatureFlag` on the Zed side.
- It has no surface for rate-limit info — Zed's only limit-style UI is driven off `token_usage.ratio()`.

Using a select dropdown lets the user pick which of three metrics they want on-screen, with the others a click away in the list. It also degrades gracefully: the 5h and Week options are hidden entirely when the SDK hasn't emitted useful data for them, so the dropdown stays tidy during normal use. This is purely additive — the data comes from the existing SDK connection (`rate_limit_event` messages plus `lastAssistantTotalUsage` / `contextWindowSize` already tracked for `usage_update`).

## Screenshots

Note: Screenshots taken combined with effort selector code present in `rohan-patra/claude-agent-acp/main`.

Screenshot 2026-04-19 at 1 33 29 PM\
_Context Limit (default)_
<br/>

Screenshot 2026-04-19 at 1 33 55 PM\
_5hr reset timer. % used will show over 85%._
<br/>

Screenshot 2026-04-19 at 1 34 03 PM\
_Hidden_
<br/>

## Limitations of this implementation

- **Weekly usage data is rarely visible.** The SDK only emits `seven_day_opus` / `seven_day_sonnet` utilization once a warning threshold has been crossed. In normal use, only `resetsAt` arrives, and we hide the Week option rather than render a mostly-empty label. Filed upstream: https://github.com/anthropics/claude-code/issues/50518.
- **No dropdown-placement control.** ACP has no field for left-align / compact rendering; that's a client-side concern. In Zed today the option sits alongside Mode / Model.
- **Updates fire on every SDK event, not on a timer.** Both `rate_limit_event` and `result` paths synchronously emit a `config_option_update`. In practice these arrive infrequently enough (once per turn / once per rate-limit transition) that no debounce is warranted — an earlier debounced version was reviewed and backed out as overengineered.
- **Countdown labels go stale while idle.** The `5h 62% 2:15` label is built on the SDK event; we don't re-push the countdown while the session sits between prompts. The minute-precision label stays close enough for the human eye for the ~minutes-to-hours window it typically covers, but a perfectly ticking countdown would need either a per-session timer or a Zed-side re-render hook.
- **Weekly label only reads Opus.** When both Opus and Sonnet utilization are present, only Opus drives the label percentage; Sonnet appears in the tooltip. This matches the user's typical concern (weekly Opus budget being scarcer) but is a preference baked into `formatWeek`.

## Future

Should the SDK start supporting sharing the required data in a hook rather than this message event, we would want to migrate to that as source of truth. Supporting a countdown on the reset timer would also be trivial to support, just wasn't warranted for this implementation.

## Disclosure

This change was largely generated with Claude Code (Opus 4.7 1M) and reviewed + tested end-to-end in Zed. The formatters in `src/usage.ts` are pure and covered by 42 unit tests in `src/tests/usage.test.ts`; 4 additional tests in `src/tests/acp-agent.test.ts` verify `rate_limit_event` → `config_option_update` push behavior.

## Test plan

- [x] `npm run build`
- [x] `npm run test:run` (227 pass; 3 pre-existing `authorization.test.ts` failures on `main` are unrelated)
- [x] Verified dropdown labels and tooltips in Zed across: initial state (Ctx only), after rate-limit warning threshold (5h appears), after weekly warning (Week appears), and with hidden selected

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -54,6 +54,7 @@ import {
   Query,
   query,
   SDKPartialAssistantMessage,
+  SDKRateLimitInfo,
   SDKUserMessage,
   SlashCommand,
 } from "@anthropic-ai/claude-agent-sdk";
@@ -73,6 +74,7 @@ import {
   toolUpdateFromEditToolResponse,
   toolUpdateFromToolResult,
 } from "./tools.js";
+import { buildUsageConfigOption, UsageRateLimitKey, UsageView } from "./usage.js";
 import { nodeToWebReadable, nodeToWebWritable, Pushable, unreachable } from "./utils.js";
 
 export const CLAUDE_CONFIG_DIR =
@@ -147,6 +149,14 @@ type Session = {
    *  DEFAULT_CONTEXT_WINDOW, refreshed from each result's modelUsage, and
    *  invalidated when the user switches the session's model. */
   contextWindowSize: number;
+  /** Which metric the "Usage" dropdown is currently displaying. */
+  usageView: UsageView;
+  /** Latest context tokens used (total) and window size, used to render the
+   *  `Ctx:` view. Null until the first assistant message reports usage. */
+  usageContext: { used: number; size: number } | null;
+  /** Latest rate-limit info keyed by `rateLimitType`. Updated from
+   *  `rate_limit_event` SDK messages. */
+  usageRateLimits: Partial<Record<UsageRateLimitKey, SDKRateLimitInfo>>;
 };
 
 /** Compute a stable fingerprint of the session-defining params so we can
@@ -735,6 +745,11 @@ export class ClaudeAcpAgent implements Agent {
                   },
                 },
               });
+              session.usageContext = {
+                used: lastAssistantTotalUsage,
+                size: session.contextWindowSize,
+              };
+              await this.pushUsageOption(params.sessionId);
             }
 
             if (session.cancelled) {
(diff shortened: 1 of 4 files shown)


rohan-patra:
great idea! although I doubt they will merge this, I have added it to my fork 😊

clean solution imo https://github.com/rohan-patra/claude-agent-acp/commit/6cf982eddb57cf46afae6b8b3f7aebc793985809

benbrandt:
Correct, not planning to merge this as this doesnt' seem like a valid use of a config option tbh. Understand this is an issue though

kayakyakr:
> Correct, not planning to merge this as this doesnt' seem like a valid use of a config option tbh. Understand this is an issue though

Totally. Think of this as more of a PoC/status-check for what we could do _today_ in regards to displaying these values (and something I'll run locally, since it fits my workflow).

Went ahead and opened a feature request in Zed itself to implement this support the "correct" way: https://github.com/zed-industries/zed/discussions/54335

Seluj78:
> **Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

Any chance of reviving this? It would unblock a downstream consumer: agent-of-empires (https://github.com/agent-of-empires/agent-of-empires) embeds claude-agent-acp in a web cockpit and wants a `/usage`-style limit meter, but limit data is local-only today and never reaches ACP clients. The display-only `Usage` SessionConfigOption here is the exact shape that would let clients render it without a spec change. What blocked the merge?

PR #569, opened 2026-04-19
feat: add /btw ephemeral side-question command

rohan-patra:
Adds `/btw <question>` — a one-shot slash command that answers a tangential question with full access to the current conversation's context but does NOT persist either the question or answer to the session transcript. Subsequent turns see a pristine context. Mirrors the Claude Code CLI's `/btw` semantics using only public SDK primitives:

- `persistSession: false` — nothing written to `~/.claude/projects/`
- `resume: <sessionId>` — side query has main-session context
- `maxTurns: 1` + `tools: []` + `disallowedTools: ["*"]` — single turn, tools disabled at the model-context level (not just execution)

The side query's text is forwarded to the client on the main ACP sessionId via `toAcpNotifications` so the answer appears inline. Serialized behind any in-flight main turn via a new `idleResolvers` list to avoid interleaved output. Interruptible via ACP `cancel()`.

Tests cover empty-argument, concurrent-guard, cancel-while-waiting-for-idle, and the available-commands append path.


josevalim:
Instead of providing this via the SDK, I'd prefer if we extended ACP to make this possible at a higher level. That would mean adding two new options: disableAllTools and persistSession? This way you could implement `/btw` in your product in a way that works across all agents. Thoughts? 🤔

rohan-patra:
@josevalim this is just a simple implementation to get to feature parity with the CC CLI. A new ACP primitive would hopefully be full thread forking or message deletion.

rohan-patra:
@benbrandt Are you guys open to this feature?

DaisukeYoda:
Adding a user voice. I use `/btw` daily in the CLI and missing it in the Zed/ACP flow is a real downgrade.

On the framing: @josevalim's `disableAllTools` + `persistSession` proposal is reasonable as a general extension, but I don't think it fits `/btw` specifically. `/btw` is one-shot (one turn, no persistence, no follow-up), while an ACP `session` carries a full lifecycle and per-session state. Flags make the session lighter but don't change its shape, so you'd still pay a `session/new`, one turn, `session/close` cycle per call.

The primitive that matches `/btw` looks closer to a one-shot side prompt on an existing session, not a new session with flags. That's a much bigger spec change.

Meanwhile this PR is well-scoped and well-tested. I'd vote to land it as an SDK-level interim and pursue the ACP abstraction separately.

PR #571, opened 2026-04-20
fix: Restore summarized thinking display on Opus 4.7

rohan-patra:
## Summary

Opus 4.7 silently changed the default for `thinking.display` from `"summarized"` to `"omitted"`. As a result, thinking blocks still stream but their `thinking` field arrives empty, so users no longer see Claude's reasoning surfaced in the ACP `agent_thought_chunk` notifications that flow from this SDK.

This PR sets `Options.thinking = { type: "adaptive", display: "summarized" }` as the default when constructing the session's SDK `Options`, restoring the behavior that existed on Opus 4.6. The default is applied before the `...userProvidedOptions` spread, so callers supplying their own `thinking` config via `_meta.claudeCode.options.thinking` still override it.

Anthropic docs:
- [Adaptive thinking](https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking) — on Opus 4.7, `adaptive` is the only supported thinking mode.
- [Controlling thinking display](https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#controlling-thinking-display) — explicitly notes that `display` must be set to `"summarized"` on Opus 4.7 to get summarized thinking back.

## Test plan

- [x] `npm run build` passes
- [x] `npm run test:run` passes (184 tests)
- [ ] Manual verification: start a session on Opus 4.7, prompt a reasoning-heavy question, confirm `agent_thought_chunk` notifications arrive with non-empty text

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -1544,6 +1544,11 @@ export class ClaudeAcpAgent implements Agent {
     const options: Options = {
       systemPrompt,
       settingSources: ["user", "project", "local"],
+      // Opus 4.7 silently changed `thinking.display` to default `"omitted"`,
+      // so thinking blocks stream with an empty `thinking` field. Opt back into
+      // summarized thinking to restore the prior Opus 4.6 behavior. Overridable
+      // via `userProvidedOptions.thinking` below.
+      thinking: { type: "adaptive", display: "summarized" },
       ...(maxThinkingTokens !== undefined && { maxThinkingTokens }),
       ...userProvidedOptions,
       env: {


josevalim:
Although I personally like the thinking summaries, I don't think we should enable it by default:

1. I believe we should stay close to the model/SDK defaults whenever possible (so Claude Code users have a consistent UX)
2. According to the linked docs on why omitting tokens: "The primary benefit is faster time-to-first-text-token when streaming: The server skips streaming thinking tokens entirely and delivers only the signature, so the final text response begins streaming sooner."

If you want to overwrite the SDK conventions, you can already enable thinknig via `sessionMeta?.claudeCode?.options`.

rohan-patra:
@josevalim ended up just adding it as an env config here https://github.com/rohan-patra/claude-agent-acp/commit/d3d38c5b1cfdc566f93f106b9721100e27e43def

benbrandt:
Was going to say I think an env var could be an option here for folks using the acp package via the registry or something

levrik:
@benbrandt Would you accept a PR for this? I'm coming from this Zed issue: https://github.com/zed-industries/zed/issues/56375

PR #597, opened 2026-04-26
fix: Fall through to next CLI candidate on ENOENT spawn

josevalim:
When bunx installs the agent, it materializes both the glibc and musl optional dependencies as resolvable packages even though only one ships the actual binary on disk. The previous resolve-time fallback returned the musl path successfully, then `spawn` failed with ENOENT at runtime.

Replace `claudeCliPath` with `spawnClaudeCli`, which spawns each resolved candidate and falls through on ENOENT, and tries glibc before musl.

Assisted by: Claude Opus 4.7


josevalim:
I'd add that I have not figured out an easy way to know if the current system is musl or not, hence the solution above.

SteffenDE:
Closing in favor of https://github.com/agentclientprotocol/claude-agent-acp/pull/598.

PR #598, opened 2026-04-27
detect glibc and use correct binary

### LINKED ISSUE #141: Claude: Failed to load models
Hi,
I had Claude working fine for quite some time, but recently encountered this issue preventing it from working. In the chat interface, I can see the error "Failed to load models". Clicking on that takes me to the Providers section of the Settings for Tidewave where I can see that Claude Code is connected. Opening the Terminal shows the error below "ENOENT: no such file or directory, posix_spawn '/tmp/bunx-1000-@agentclientprotocol/claude-agent-acp@0.31.0/node_modules/@anthropic-ai/claude-agent-sdk-linux-x64-musl/claude'". But I can start Claude fine from the terminal as shown below. 

Image

I am running the latest version of Tidewave, as well as the latest version of Claude. The issue looks similar to this one: https://github.com/tidewave-ai/tidewave_app/issues/120. But I have tried deleting these folders and restarting without any change in behaviour:

> rm -rf /tmp/bun-node-30e609e08/
> rm -rf /tmp/bunx-1000-@agentclientprotocol/
> rm -rf /home/antonyb/.cache/tidewave/

SteffenDE:
On Linux, the current code prefers the musl binary if multiple versions are installed, but it actually crashes on glibc systems.

Supersedes https://github.com/agentclientprotocol/claude-agent-acp/pull/597.

Importantly, we need to pass the detected binary to `pathToClaudeCodeExecutable`, otherwise the SDK itself selects the wrong version.

See https://github.com/tidewave-ai/tidewave_app/issues/141.
Closes https://github.com/tidewave-ai/tidewave_app/issues/141.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -274,12 +274,22 @@ export async function claudeCliPath(): Promise<string> {
   const { createRequire } = await import("node:module");
   const req = createRequire(import.meta.resolve("@anthropic-ai/claude-agent-sdk"));
   const ext = process.platform === "win32" ? ".exe" : "";
+  // On linux, both glibc and musl variants may be installed side-by-side
+  // (e.g. bunx hydrates every optional dep), so picking one by trial is
+  // unreliable: the wrong binary segfaults at runtime instead of failing to
+  // spawn. Detect the runtime libc and prefer the matching variant, falling
+  // back to the other only if the preferred one isn't installed.
   const candidates =
     process.platform === "linux"
-      ? [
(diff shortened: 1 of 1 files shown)


josevalim (approved):

SteffenDE:
@benbrandt if you don't see any issues with this, it would be great if we can ship a new release that includes it soon. I can confirm that this fixes starting sessions on my Fedora Linux installation.

benbrandt (approved):
Thanks I tried to copy the SDK behavior which is why you are seeing different behavior without passing it in , btu I was a bit skeptical about the appraoch they took

PR #601, opened 2026-04-27
fix: tolerate invalid values in settings.json

AlexPl292:
A broken value in settings.json (e.g. an unknown permissions.defaultMode or a non-string model) used to throw and prevent session creation. Match the Claude Code SDK behavior by logging the bad value and falling back to defaults instead, so a single typo no longer renders the agent unusable.


benbrandt (approved):
I pushed the logging/defaults into the functions themselves, rather than add try/catch on the outside. 

Thanks!

PR #617, opened 2026-04-30
fix: surface Bash tool image output instead of dropping it

davethedev3:
## What's broken

The Bash branch of `toolUpdateFromToolResult` (`src/tools.ts`) guards the array-content path on `"text" in result[0]`. When the local Bash tool produces image content — `[{ type: "image", source: { type: "base64", ... } }]`, e.g. when a command pipes a base64 data URI — the guard rejects the array and `output` stays empty. The image is silently dropped from the ACP `tool_call_update` and the client renders nothing.

## Fix

Distinguish text-only arrays (existing terminal-stream behavior) from arrays containing non-text blocks. For the latter, bypass the terminal `_meta` channel (binary payloads can't be streamed through it) and delegate to `toAcpContentUpdate` so images surface as ACP content blocks.

The change is contained to the Bash case in `toolUpdateFromToolResult`; no other tool dispatch is affected.

## Tests

Three regression tests added in `src/tests/tools.test.ts` under a new `describe("with image array tool_result …")`:

- pure image, `supportsTerminalOutput: false` — image surfaces as ACP content
- pure image, `supportsTerminalOutput: true` — `_meta` is correctly bypassed and image still surfaces
- mixed text + image — both blocks surface in order

All existing tests pass. Total: 240 → 243.

## Validation

- `npm run check` ✅
- `npm run build` ✅
- `npm run test:run` ✅ (243 passed, 13 integration skipped)


benbrandt (approved):
thanks!

PR #646, opened 2026-05-11
Allow launchers to skip initial Claude setModel

imwalkinhere:
## Summary

- add a launcher-controlled escape hatch for the initial `query.setModel(...)` call when the model is already baked into the Claude subprocess environment
- treat both `CLAUDE_AGENT_ACP_SKIP_INITIAL_SET_MODEL=1` and the existing OpenClaw-facing `CLAUDE_ACP_BAKE_MODEL=1` marker as signals to skip the redundant setModel call
- cover the behavior with a settings test that selects the current model from `ANTHROPIC_MODEL` without calling `setModel`

## Context

Some ACP launchers select the Claude model before process startup via environment variables. In that mode, immediately replaying `setModel` after initialization can hit provider/runtime paths that are more fragile than launch-time model selection. This keeps the default behavior unchanged unless a launcher explicitly opts into the skip marker.

## Tests

- `npm run test:run -- src/tests/acp-agent-settings.test.ts`
- `npm run build`
- `npm run format:check`
- `npm run lint`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -2439,7 +2439,12 @@ async function getAvailableModels(
     }
   }
 
-  await query.setModel(currentModel.value);
+  if (
+    process.env.CLAUDE_AGENT_ACP_SKIP_INITIAL_SET_MODEL !== "1" &&
+    process.env.CLAUDE_ACP_BAKE_MODEL !== "1"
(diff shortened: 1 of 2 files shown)


benbrandt:
Addressed in #704

PR #647, opened 2026-05-11
feat: add 'Default' option to effort level config

AceGain:
Add a "Default" selection to the effort level config option, allowing users to let the SDK decide the effort level instead of forcing a specific level. When "default" is selected, applyFlagSettings is called with effortLevel: undefined, clearing any previous override.

Replaces the previous fallback logic (xhigh -> high -> supportedLevels[0]) with a simpler "default" fallback that lets the SDK use its own default.


AceGain:
Please,agree it. The third-party model may not support the 'Effort', or not support 'high' and 'xhigh'. We need it.

AceGain:
@cla-bot check

PR #653, opened 2026-05-13
feat: add agent selection as ACP session config option

dsturnbull:
Expose available Claude Code agents (user + plugin) as a selectable config option in the ACP session response. Zed renders this as a dropdown automatically alongside Mode, Model, and Effort.

When the user switches agents:
- Aborts the current subprocess
- Resumes the session with the new agent (preserving history)
- Falls back to a fresh session if no history exists yet

Agent discovery uses the SDK's supportedAgents() API which returns all agents (user ~/.claude/agents/*.md + plugin agents), with a filesystem fallback for offline discovery.


benbrandt:
It looks like we can support this wihtout killing the process now: https://github.com/anthropics/claude-agent-sdk-typescript/blob/main/CHANGELOG.md#03161

benbrandt:
handled here: https://github.com/agentclientprotocol/claude-agent-acp/pull/794

PR #691, opened 2026-05-19
fix: Skip stale idle left buffered after a failed turn

0x5ea000000:
## Problem

When a prompt fails mid-turn (e.g. a 400 rate-limit error from a result with `is_error=true`), the SDK's trailing `session_state_changed{idle}` turn-boundary message stays buffered in `session.query`. The next prompt's loop immediately consumes that stale idle and returns with no content — shifting every subsequent user response one turn late until the session is restarted https://github.com/zed-industries/zed/issues/53483

## Root cause

The prompt while-loop exits via `throw` before it reaches the `session_state_changed{idle}` message that the SDK always emits as a turn boundary. That idle sits in the query buffer and the next invocation of `prompt()` hits it first, taking the normal "end of turn" return path with no content.

## Fix

Rather than consuming the idle in the `finally` block (which advances the SDK `AsyncGenerator` and has a side effect: the CLI stops filtering the failed turn from the next API request, causing context from the failed message to bleed into later responses), we instead set a `session.staleIdlePending` flag.

When the next `prompt()` call starts and its loop encounters the idle at the front of the queue, it checks the flag. If set, it clears the flag and `break`s — skipping past the stale boundary and continuing to read the real response — rather than returning early.

Release Notes:

- Fixed responses being offset by one turn after an agent prompt error (e.g. a rate-limit 400), requiring a session restart to recover


benbrandt:
Closing in favor of #706

PR #707, opened 2026-05-21
fix: Remove hide Claude auth flag handling

benbrandt:
Now that the terms are clear, we can let people use their credits if
they like in all scenarios.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -330,10 +330,6 @@ function isMuslLibc(): boolean {
   return !report?.header?.glibcVersionRuntime;
 }
 
-function shouldHideClaudeAuth(): boolean {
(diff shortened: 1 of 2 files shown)


AlexandrSuhinin:
I'm not sure if it's safe to remove this yet. I still see a restriction in the official docs: https://code.claude.com/docs/en/agent-sdk/overview#get-started
image

Also, the change will be applied after June 15, so it's probably too early for the fix: https://support.claude.com/en/articles/15036540-use-the-claude-agent-sdk-with-your-claude-plan

benbrandt:
Ok doesn't matter to me. Can sit here

AlexandrSuhinin (approved):
@benbrandt I have an update: we're fine with removing this argument now.

PR #729, opened 2026-05-29
Fix potential ReDoS in local-command metadata stripping

### LINKED ISSUE #727: Potential ReDoS in local-command metadata stripping regex
## Summary

`src/acp-agent.ts` currently strips local-command metadata with a regular expression that can be flagged by CodeQL as a polynomial-time regex on uncontrolled input.

Current code on `main`:

```ts
const LOCAL_COMMAND_TAG_PATTERN =
  /<(command-name|command-message|command-args|local-command-stdout|local-command-stderr)>[\s\S]*?<\/\1>/g;

function stripMarkerTags(text: string): string {
  return text.replace(LOCAL_COMMAND_TAG_PATTERN, "");
}
```

GitHub Advanced Security reports this as:

> CodeQL / Polynomial regular expression used on uncontrolled data

## Why This Matters

The input being processed here comes from persisted/replayed session content. Even though the metadata format is expected to be internal and well-formed, replay paths can encounter malformed or adversarial-looking text.

A long string containing many marker-like fragments, or an unterminated marker, should not require regex backtracking to decide whether anything should be stripped.

## Suggested Fix

Replace the regex-based implementation with a small linear scanner that:

- only strips known complete metadata tags,
- leaves unterminated tags unchanged,
- preserves normal prose mixed with metadata,
- avoids regex backtracking on malformed input.

For example:

```ts
const LOCAL_COMMAND_TAG_NAMES = [
  "command-name",
  "command-message",
  "command-args",
  "local-command-stdout",
  "local-command-stderr",
] as const;

function stripMarkerTags(text: string): string {
  let stripped = "";
  let cursor = 0;

  while (cursor < text.length) {
    const openStart = text.indexOf("<", cursor);
    if (openStart === -1) {
      stripped += text.slice(cursor);
      break;
    }

    const tagName = LOCAL_COMMAND_TAG_NAMES.find((name) =>
      text.startsWith(`<${name}>`, openStart),
    );

    if (!tagName) {
      stripped += text.slice(cursor, openStart + 1);
      cursor = openStart + 1;
      continue;
    }

    const openEnd = openStart + tagName.length + 2;
    const closeTag = `</${tagName}>`;
    const closeStart = text.indexOf(closeTag, openEnd);

    if (closeStart === -1) {
      stripped += text.slice(cursor);
      break;
    }

    stripped += text.slice(cursor, openStart);
    cursor = closeStart + closeTag.length;
  }

  return stripped;
}
```

## Suggested Regression Tests

```ts
it("leaves unterminated marker tags unchanged", () => {
  const text = "<command-args>opus";
  expect(stripLocalCommandMetadata(text)).toBe(text);
});

it("handles many marker-looking fragments without dropping real prose", () => {
  const markerLikeNoise = "<command-args".repeat(5000);
  expect(stripLocalCommandMetadata(`${markerLikeNoise}hi`)).toBe(`${markerLikeNoise}hi`);
});
```

## Notes

This appears to have been introduced by the session replay/local-command metadata handling change, not by the later Claude Agent SDK dependency bump.

Related PRs:

- https://github.com/agentclientprotocol/claude-agent-acp/pull/593
- https://github.com/agentclientprotocol/claude-agent-acp/pull/722

If maintainers prefer this to be handled as a security advisory rather than a public issue, I am happy to move the report there.

jsgrrchg:
Closes #727.

This replaces the local-command metadata stripping regex with a deterministic scanner over the known SDK local-command marker tags.

The previous implementation used a regex over transcript content to remove marker blocks such as `<command-name>...</command-name>` and `<local-command-stdout>...</local-command-stdout>`. CodeQL flags that pattern as a potential polynomial-time regex on uncontrolled input.

The new implementation avoids matching arbitrary content with a regex. It scans only for the five known local-command marker tags and strips complete, well-formed marker blocks while preserving malformed or literal marker-like text.

Covered behavior:

- strips valid SDK local-command metadata blocks
- preserves real prose around stripped marker blocks
- preserves unknown marker-like tags
- preserves incomplete or mismatched marker-like tags
- avoids pairing an incomplete opening tag with a later valid block
- handles long malformed marker-like input and unterminated fragments without dropping text

This is intended as a defensive hardening change for the CodeQL finding rather than a claim of a confirmed exploit.

Tests:

- `npm run test:run -- src/tests/acp-agent.test.ts`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -347,11 +347,116 @@ const LOCAL_ONLY_COMMANDS = new Set(["/context", "/heapdump", "/extra-usage"]);
 // payload in these XML-like markers that the CLI uses for its own display.
 // The live prompt loop drops them; replay must strip them too or they leak
 // into the UI on session/load.
-const LOCAL_COMMAND_TAG_PATTERN =
-  /<(command-name|command-message|command-args|local-command-stdout|local-command-stderr)>[\s\S]*?<\/\1>/g;
+const LOCAL_COMMAND_TAGS = [
+  "command-name",
+  "command-message",
+  "command-args",
+  "local-command-stdout",
+  "local-command-stderr",
+] as const;
+
+type LocalCommandTag = (typeof LOCAL_COMMAND_TAGS)[number];
+type LocalCommandTagToken = {
+  marker: string;
+  tag: LocalCommandTag;
+  closing: boolean;
+};
+type LocalCommandTagPositions = {
+  openings: number[];
+  openingHead: number;
+  closings: number[];
+  closingHead: number;
+};
+
+const LOCAL_COMMAND_TAG_TOKENS: LocalCommandTagToken[] = LOCAL_COMMAND_TAGS.flatMap((tag) => [
+  { marker: `<${tag}>`, tag, closing: false },
+  { marker: `</${tag}>`, tag, closing: true },
+]);
+
+function readLocalCommandTagToken(text: string, index: number): LocalCommandTagToken | null {
+  for (const token of LOCAL_COMMAND_TAG_TOKENS) {
+    if (text.startsWith(token.marker, index)) return token;
+  }
+  return null;
+}
 
 function stripMarkerTags(text: string): string {
-  return text.replace(LOCAL_COMMAND_TAG_PATTERN, "");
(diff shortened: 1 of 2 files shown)


jsgrrchg:
Closing this PR since upstream preferred a simpler solution., https://github.com/agentclientprotocol/claude-agent-acp/pull/738

Issue #459, opened 2026-03-23
Feature request: expose remote-control (/rc) via ACP


mambucodev:
The `claude --remote-control` / `/rc` flag was added as a research preview recently and lets you attach a running Claude Code session to claude.ai/code from your browser or phone.

Right now it's CLI-only — if you start a session from Zed's sidebar (via ACP) there's no way to attach remotely to it. Would be great if the ACP adapter could either:

- pass `--remote-control` when starting a session (opt-in flag or setting), or
- expose a way to get the remote URL for an already-running session

Not sure where the right hook is on the ACP side, but the use case is pretty clear: you start something long-running from Zed, step away from your desk, and want to check in or steer it from claude.ai/code without needing a terminal open.

Happy to test if there's anything experimental.

mchisolm0:
I can see why that would be helpful. It sounds like from <https://github.com/agentclientprotocol/claude-agent-acp/issues/23#issuecomment-3255432769>, some of the slash commands do not work well with the Claude Code SDK and had to be blocklisted. Not sure if the `/rc` command is one of those, though.

Here is the text from that comment:
> Unfortunately, many built-in commands do not work well with the Claude Code SDK; some produce no output, others get the conversation into a broken state, and many other variations exist, so we had to blocklist them.
>
> This is out of our hands, so please open issues in [anthropics/claude-code](https://github.com/anthropics/claude-code) for each command you'd like to use.

Sergio-3:
Some evidence on the two open questions here, gathered against the current release (adapter `0.62.0`, which pins `@anthropic-ai/claude-agent-sdk` `0.3.219`):

**1. `/rc` is not blocked by the SDK slash-command blocklist.** The capability is already exposed as a control request on the runtime `Query` object:

```js
async enableRemoteControl(e, t) {
  return (await this.request({ subtype: "remote_control", enabled: e, ...(t !== void 0 && { name: t }) })).response
}
```

It returns the session URL, so nothing needs to be rendered in a terminal for the bridge to come up. What *is* terminal-only is the CLI's `local-jsx` UI for `/remote-control` — which is also why the command never appears in `supportedCommands` over ACP. The underlying toggle itself is plain RPC and works headless.

[...]

PR #735, opened 2026-06-01
feat: add /remote-control command over ACP

Dhananjay-JSR:
Adds `/remote-control` (alias `/rc`) to surface Claude Code [Remote Control](https://code.claude.com/docs/en/remote-control) over ACP — connect a running session to claude.ai/code and continue it from a phone or browser. Intercepts the command in `prompt()`, calls `query.enableRemoteControl(enabled, name)`, posts the returned session URL back to the client, and advertises both commands so clients (e.g. Zed) forward them. A second invocation disconnects.

Note: `enableRemoteControl` exists on the runtime `Query` (claude-agent-sdk 0.3.x) but isn't in the published types yet, so it's reached via a small type shim.

Verification: `npm run check`, `npm run build`, `npm run test:run` (new unit tests cover the toggle, error path, and command advertising). Also verified end-to-end over ACP against a real Claude binary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,6 +15,7 @@ This tool implements an ACP agent by using the official [Claude Agent SDK](https
 - Interactive (and background) terminals
 - Custom [Slash commands](https://docs.anthropic.com/en/docs/claude-code/slash-commands)
 - Client MCP servers
+- [Remote Control](https://code.claude.com/docs/en/remote-control) (`/remote-control`)
 
 Learn more about the [Agent Client Protocol](https://agentclientprotocol.com/).
 
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -482,6 +482,9 @@ type Session = {
   /** Accumulated task list for the session, keyed by task ID. Task IDs are
    *  per-session, so this state must not be shared across sessions. */
   taskState: TaskState;
+  /** Whether a Remote Control bridge is currently active for this session, so
+   *  `/remote-control` toggles connect/disconnect. */
+  remoteControlActive?: boolean;
   /** Last session title we pushed to the client via `session_info_update`.
    *  The SDK auto-generates a title in a background task and persists it to the
    *  session file; we poll it on each turn-end (`session_state_changed: idle`)
@@ -630,6 +633,19 @@ type Session = {
    *  NOT READ YET — recorded now so the mapping exists if/when we wire up
    *  fork/rewind. */
   messageIdToUuid: Map<string, string>;
+};
+
+/** Result of the SDK's `remote_control` control request. The method exists on
+ *  the runtime `Query` object (since claude-agent-sdk 0.3.x) but is not yet in
+ *  its published type definitions, so we narrow to it via this shim. */
+type RemoteControlResponse = {
+  session_url?: string;
+  connect_url?: string;
+  environment_id?: string;
+};
+
+type QueryWithRemoteControl = Query & {
+  enableRemoteControl(enabled: boolean, name?: string): Promise<RemoteControlResponse | undefined>;
 };
 
 /** Result-message origin kinds that mark an AUTONOMOUS cycle — work the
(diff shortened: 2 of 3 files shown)


Sergio-3:
Posted some verification in #459: the `enableRemoteControl` control request this PR drives does exist on the runtime `Query` object in the SDK version the current release pins (`0.3.219`), and it isn't affected by the slash-command blocklist — so the approach here looks sound rather than blocked upstream. The one thing that needs a maintainer ruling is that the method is absent from the published `sdk.d.ts`, hence the `QueryWithRemoteControl` cast.

This has had no review since it was opened (2026-06-01) and now conflicts with `main`. @Dhananjay-JSR are you still up for a rebase if a maintainer greenlights the runtime-cast approach? Happy to test a rebased build against Zed either way.

Dhananjay-JSR:
> Posted some verification in #459: the `enableRemoteControl` control request this PR drives does exist on the runtime `Query` object in the SDK version the current release pins (`0.3.219`), and it isn't affected by the slash-command blocklist — so the approach here looks sound rather than blocked upstream. The one thing that needs a maintainer ruling is that the method is absent from the published `sdk.d.ts`, hence the `QueryWithRemoteControl` cast.
> 
> This has had no review since it was opened (2026-06-01) and now conflicts with `main`. @Dhananjay-JSR are you still up for a rebase if a maintainer greenlights the runtime-cast approach? Happy to test a rebased build against Zed either way.

Check again

Sergio-3:
Thanks for the 27 Jul push — I should have acknowledged it here sooner.

Where it stands against current `main` (`08a62ed`): your rebase landed on `v0.62.0` (`53a0c36`), and `main` has moved 12 commits since, which is why GitHub still reports the branch as conflicting. Merging your head into `main` today auto-merges `README.md` and `src/acp-agent.ts` — the only conflict left is `src/tests/acp-agent.test.ts`, where `main` has since gained tests from #930, #923, #919 and #916.

I've been running this branch against Zed for the past week and it works. For maintainers, the design call I see is whether the type assertion used to reach `enableRemoteControl` is acceptable while the method is missing from the published types (absent from `sdk.d.ts` in 0.3.219).

Two gaps I'd suggest landing with the command — without them you get an answer with no visible question:

**1. Enabling the bridge doesn't start the stream consumer.** The consumer is lazily started on the first `prompt()`; this PR intercepts the command before entering the turn queue and returns, so a session that enables the bridge as its first action has nothing draining the stream, and the bridge's frames go unread until an unrelated prompt starts the consumer. `ensureConsumer(session, sessionId)` on enable closes this half — that's what my local build does — but it isn't the whole fix: once the frames actually arrive, they hit (2).

**2. The remote message arrives as a replay and is dropped.** The CLI does put it on the stream, flagged `isReplay: true`, and the consumer's unmatched-replay guard drops it — deliberately so, for turn echoes; remote-origin replays just need distinguishing from those. Its answer streams normally, which is what makes the asymmetry easy to miss.

Measured with a small SDK host configured like this adapter (`extraArgs: {"replay-user-messages": ""}`), enabling the bridge and then sending one message from claude.ai/code with no local input at all (condensed from the full stdout; frame order preserved):

```
13:50:33  system/bridge_state state=ready
13:50:33  system/bridge_state state=connected
          … no local input for the next 1m45s …
13:52:19  system/session_state_changed · command_lifecycle ×2 · system/init
13:52:23  user       isReplay=true  origin={"kind":"human"}  "Hi again, test 2"
13:52:23  assistant  (reply text elided)
13:52:23  rate_limit_event · system/post_turn_summary · result
13:52:23  command_lifecycle · system/session_state_changed
```

<details><summary>the host used for that measurement</summary>

```js
import { query } from "@anthropic-ai/claude-agent-sdk";
async function* idleInput() { await new Promise((r) => setTimeout(r, 300_000)); }
const session = query({
  prompt: idleInput(),
  options: { cwd: process.cwd(), extraArgs: { "replay-user-messages": "" } },
});
setTimeout(async () => {
  const res = await session.enableRemoteControl(true, "repro");
  console.error("open:", res?.session_url);
}, 4000);
for await (const message of session) console.log(JSON.stringify(message));
```

In a run without the flag, the `user` frame didn't appear — worth knowing if anyone reproduces this outside the adapter.
</details>

One detail for whoever implements the mirroring: `origin` doesn't mark the message as remote — it arrives as `{"kind":"human"}`, the same value a locally typed message carries in the session transcript. What works is tracking the uuids this adapter pushed, including `steer()`'s, since those create no `Turn` and so match no queued turn either.

@Dhananjay-JSR I have the test-file resolution plus both gaps on a local branch. Happy to send it as a patch to cherry-pick, or open a PR against your branch — whichever you prefer. If maintainers would rather see it as one PR here, I can do that instead.

FrameAutomata:
Question for maintainers, since this seems to be the open design call rather than anything in the code:

`enableRemoteControl` exists on the runtime `Query` but not in the published types. I checked 0.3.247 (latest) — still `sdk.mjs` only, absent from `sdk.d.ts` — so "wait for the types to land" has no visible ETA.

Is the type assertion acceptable here, or would you rather this wait until the method is published?

Asking because I have this rebased on current `main` and would rather not keep it current against a decision that hasn't been made.

FrameAutomata:
Re-pinging this. The 27 Aug question is still the only thing blocking, and it's a one-line ruling rather than a code change.

`enableRemoteControl` is still runtime-only. I re-checked today against `0.3.257` (what `main` pins) and `0.3.260` (latest): present in `sdk.mjs`, absent from `sdk.d.ts` in both. So nothing has moved since 0.3.247, and "wait for the published types" has no visible ETA.

The call I need: is the `QueryWithRemoteControl` cast acceptable as an interim, or should this wait until the method is published in the types? Either answer works — I'd just rather not keep the branch current against a decision that hasn't been made.

@Ololoshechkin you've landed most of the recent PRs here — would you be able to make the call, or point me at whoever should?

For context, I have this rebased and working against Zed, plus a follow-up commit that fixes delivery of Remote Control peer messages over ACP. The branch is ~13 commits behind `main` now; I'll refresh it as soon as there's a direction.

PR #750, opened 2026-06-05
feat: Send message ids for assistant and user messages

benbrandt:


SteffenDE:
Clever, I didn't think about restoring the mapping for fork/resume. Very nice!

benbrandt:
Yeah I went through several rounds validating _both_ of these ids are available... sure would be nice if we should just use one 😅

PR #755, opened 2026-06-09
feat: bridge AskUserQuestion through ACP elicitation

doppleware:
When the connected ACP client advertises form-mode elicitation via `clientCapabilities.elicitation.form`, wire two SDK callbacks so user-input flows actually work over ACP:

- `onElicitation` forwards MCP-server-initiated elicitations through `client.unstable_createElicitation`. ACP and MCP elicitation share a near-identical form/url + accept/decline/cancel shape, so this is a thin pass-through with a session-scope envelope.
- `onUserDialog` handles CLI-initiated dialogs. For `dialogKind: "askUserQuestion"` each question is mapped to a single-property form schema (`{ answer: { type: "string", oneOf: [...options] } }` for single-select, an array schema for multiSelect) and the user's pick is returned to the SDK as the dialog result. Unknown dialogKinds answer `cancelled` per the SDK contract.

`AskUserQuestion` is removed from `disallowedTools` only when the client advertises elicitation; otherwise we keep the existing behavior so the built-in (which expects an interactive terminal) doesn't run blind.

Bridge helpers live in their own module (`src/elicitationBridge.ts`) so the wiring in `acp-agent.ts` stays a few lines.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -43,11 +43,15 @@ import {
 import {
   CanUseTool,
   deleteSession,
+  ElicitationRequest as SdkElicitationRequest,
+  ElicitationResult as SdkElicitationResult,
   getSessionMessages,
   listSessions,
   McpServerConfig,
   ModelInfo,
   ModelUsage,
+  OnElicitation,
+  OnUserDialog,
   Options,
   PermissionMode,
   PermissionUpdate,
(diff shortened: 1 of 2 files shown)


benbrandt:
Happened to do this yesterday. Unfortunately seems to have ot go through canuseTool which wasn't intuitive. Merged in #756

Issue #297, opened 2026-02-08
Enabling extended thinking


anilzeybek:
I would like to use Claude Code with extended thinking enabled. However, it appears that this feature is not enabled by default, and I cannot find a configuration setting to change it.

Could you please add an option to enable this feature?

sha2fiddy:
+1 I am not able to get Claude's Extended Thinking to work from the Zed agent panel. It works fine from the terminal.

josevalim:
You can do this by setting the `MAX_THINKING_TOKENS` environment variable. Claude Code defaults to 31999. You can also pass it as an option when stasrting ACP (if you are using the protocol directly).

sha2fiddy:
Thanks @josevalim, but I believe `MAX_THINKING_TOKENS` approach has been depreciated. It should be configured with `CLAUDE_CODE_EFFORT_LEVEL=high` (see https://code.claude.com/docs/en/model-config#adjust-effort-level).

I spent a good while with Claude changing these variables, as well as any relevant settings I could find in Zed.  No success getting the Claude Agent to use extended thinking from the agent panel.

anilzeybek:
Since ACP integration uses Claude Code under the hood, shouldn't these flags be configured for the Claude Agent by default?

sha2fiddy:
fwiw, I do have this set as my default, and that works as expected everywhere else I use Claude Code.  Just not in this Zed integration.

gayanper:
Is this wrapper using agent sdk? If so the agent sdk provide supported effort levels as part of model info i think. Then it can be set in the system prompt. So may be we need some changes. Similar to how opencode shows model-effort for models that support efforts.

kayakyakr:
The need for this has become more important:

https://x.com/bcherny/status/2029970236460691885

Anthropic dropped the default effort from high to medium and since the ACP wrapper only launches the default model, you only ever get medium now.

PR #766, opened 2026-06-12
fix: Empty thinking text in thought chunks on current Claude models

pavelpenev:
Recent Claude models default the API's `thinking.display` field to "omitted". Thinking blocks still stream, but signature-only: the block opens, a single signature_delta arrives, and no thinking_delta ever carries text. The agent then emits agent_thought_chunk notifications whose text is always empty, so ACP clients render blank thought sections — regardless of effort level or MAX_THINKING_TOKENS.

Request `display: "summarized"` explicitly so thought chunks carry visible summary text again:

- unset MAX_THINKING_TOKENS → {type: "adaptive", display: "summarized"} (adaptive matches the SDK default on current models, but is now sent explicitly where previously no thinking config was sent at all)
- positive budget → display added to the enabled config
- 0 (disabled) unchanged

A thinking config provided via _meta.claudeCode.options still overrides this default wholesale, so embedders retain full control.

Verified over ACP traffic with an Emacs agent-shell client on claude-fable-5, claude-opus-4-8, and claude-sonnet-4-6: before, every agent_thought_chunk had empty text; after, summarized reasoning streams and renders.

Related: #297

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -3880,20 +3880,27 @@ async function fetchContextUsedTokens(query: Query, logger: Logger): Promise<num
 /** Translate the legacy `MAX_THINKING_TOKENS` env var into the SDK's `thinking`
  *  option. The `maxThinkingTokens` option it used to feed is deprecated and
  *  reduced to on/off on current models, so map the value to explicit thinking
- *  config instead: unset → `undefined` (SDK default, adaptive on models that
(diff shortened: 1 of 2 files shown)


benbrandt:
We previously decided that we don't want to change the default here as Anthropic changed the default for a reason, but I added a fix to skip empty thinking blocks: https://github.com/agentclientprotocol/claude-agent-acp/pull/793

PR #789, opened 2026-06-18
fix: duplicate assistant text when turn activates mid-message

hugobarauna:
I generated this PR using AI. if not helpful, please delete it.

Or, maybe its value is just to show a potential bug.

I found this bug when using Tidewave. I noticed text was being duplicated when using claude and claude-agent-acp 0.47.0.

I asked the agent to investigate, and this PR seems to fix it.

---

## Summary

- Stop `resetTurnScratch()` from nulling `currentStreamMessageId` on turn activation, which was causing the assistant's message text to render **twice** (concatenated) in the feed.

## Problem

When a turn activates mid-message, the agent's message text renders twice, back-to-back. Only prose duplicates — never tool calls — and specifically the text/thinking block that streams *after* the activation. Reproduces on `0.47.0` with a prompt like *"say a sentence, run a shell command that sleeps 10s, then say a sentence."*

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -972,7 +972,12 @@ export class ClaudeAcpAgent implements Agent {
       lastAssistantError = undefined;
       lastRefusalExplanation = null;
       compactionInProgress = false;
-      currentStreamMessageId = undefined;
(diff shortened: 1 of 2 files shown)


benbrandt (commented):
Thanks!

PR #797, opened 2026-06-19
fix: suppress default context usage updates

### LINKED ISSUE #596: usage_update.size reports 200000 for default modelId even though it resolves to the 1M Opus variant
> **Provenance:** This issue was drafted by analyzing the `claude-agent-acp@0.31.0` source and ACP traffic captured in production. The findings were reviewed and verified by a human (the issue reporter) before filing — not an unfiltered AI summary.

## Summary

When an ACP client uses the `default` modelId served by `claude-agent-acp`, every `session/update` `usage_update` carries `size: 200000`, even though `default` is described in the same `session/new` payload as **"Opus 4.7 with 1M context"**. The expected size is `1_000_000`. This causes the client UI to render usage like `332k / 200k` (used exceeding displayed limit) for any Opus-1M session.

## Reproduction

1. Connect any ACP client to `claude-agent-acp@0.31.0`.
2. Issue `session/set_config_option { configId: "model", value: "default" }`. The agent acks; subsequent prompts use the 1M Opus variant.
3. Send a prompt that grows context past 200000 tokens.
4. Observe `session/update` `usage_update` notifications: `size: 200000` throughout the entire turn, regardless of how much context is actually consumed.

Captured evidence from a client fixture log:

```
{
  "used": 27996,
  "size": 200000      // ← session is on opus[1m]; should be 1_000_000
}
```

For comparison, OpenCode's ACP adapter correctly reports `size: 1048576` (1M) for its 1M-context selections in equivalent fixtures.

## Root cause

`inferContextWindowFromModel` (`src/acp-agent.ts:2682-2685`) uses regex `\b1m\b`, which matches `sonnet[1m]` but does not match `default`:

```ts
function inferContextWindowFromModel(model: string): number | null {
  if (/\b1m\b/i.test(model)) return 1_000_000;
  return null;
}
```

The `set_config_option` handler (`src/acp-agent.ts:1552`) passes the literal modelId through inference:

```ts
session.contextWindowSize = inferContextWindowFromModel(value) ?? DEFAULT_CONTEXT_WINDOW;
```

For `value === "default"`, inference returns `null` → falls back to `DEFAULT_CONTEXT_WINDOW = 200_000`.

The streaming `message_start` path (`src/acp-agent.ts:940-953`) inspects the SDK-reported `message.event.message.model` string (e.g. `claude-opus-4-1-20250805`). 1M API access is enabled via beta headers, not by changing the model identifier, so this string also lacks `1m` and inference fails the same way.

The only path that produces a correct value is the `result`-message correction at `src/acp-agent.ts:846-848`, which depends on `lastAssistantModel` being non-null and on `message.modelUsage` containing the `contextWindow`. The comment at lines 842-845 already acknowledges this gap for turns with no top-level assistant message (subagent-only / sidechain-only turns).

## Impact

- First-turn UI shows `... / 200k` during the entire stream for any user on `default` model, until the `result` correction arrives.
- `used > size` ratios cause client UIs to mis-render (e.g. red over-quota indicators).
- Subagent / sidechain-only turns may never receive the correction.

## Suggested fix

Maintain a modelId-alias map alongside the regex so semantic aliases like `default` are recognized:

```ts
const MODEL_ID_ALIASES: Record<string, number> = {
  default: 1_000_000,        // currently maps to Opus 4.7 1M
  // extend as Anthropic ships new aliases
};

function inferContextWindowFromModel(model: string): number | null {
  if (MODEL_ID_ALIASES[model] != null) return MODEL_ID_ALIASES[model];
  if (/\b1m\b/i.test(model)) return 1_000_000;
  return null;
}
```

Alternative: run inference against the model **description** text (which already says "1M context") rather than the modelId. The description is authored by `claude-agent-acp` itself and changes atomically with the alias mapping. Less robust than an explicit map, but tied to a single source of truth.

Either fix preserves the existing SDK-`result.modelUsage`-driven correction path and only patches the gap during the first turn's streaming and for sidechain-only sessions.

## Note on protocol

Per ACP `v0.11.2`, `ModelInfo` has no structured `contextWindow` field. Adding one upstream to ACP would be the cleanest long-term fix: agents declare each model's context window directly, clients display it without heuristics. Until then, the alias map is the lowest-risk patch in `claude-agent-acp`.

jsgrrchg:
## Summary

Suppress live `usage_update` notifications while the session context window is still the generic `200000` fallback placeholder.

## Details

This keeps accumulating streamed usage internally, but avoids publishing a misleading live `size` value until the context window is known from either a model-name heuristic or authoritative `result.modelUsage`.

Final `result` usage updates are unchanged, so clients still receive the correct usage, cost, and context window once `modelUsage.contextWindow` is available.

Fixes https://github.com/agentclientprotocol/claude-agent-acp/issues/596

## Testing

- `npm test -- --run src/tests/acp-agent.test.ts -t "usage_update computation"`
- `npm run build`
- `npm run test:run`
- `npx prettier --check src/acp-agent.ts src/tests/acp-agent.test.ts src/tests/session-config-options.test.ts`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -259,6 +259,10 @@ type Session = {
    *  DEFAULT_CONTEXT_WINDOW, refreshed from each result's modelUsage, and
    *  invalidated when the user switches the session's model. */
   contextWindowSize: number;
+  /** True while contextWindowSize is still the generic fallback placeholder
+   *  rather than a model-name heuristic or the authoritative window reported by
+   *  modelUsage. Mid-stream usage_update events are suppressed while this holds. */
+  contextWindowSizeIsDefault: boolean;
   /** Accumulated task list for the session, keyed by task ID. Task IDs are
    *  per-session, so this state must not be shared across sessions. */
   taskState: TaskState;
@@ -1538,6 +1542,7 @@ export class ClaudeAcpAgent {
             // leave the next prompt's mid-stream updates reporting 200k.
             if (matchingModelUsage) {
               session.contextWindowSize = matchingModelUsage.contextWindow;
+              session.contextWindowSizeIsDefault = false;
             }
 
             // Send usage_update notification
@@ -1713,10 +1718,11 @@ export class ClaudeAcpAgent {
                   // Model switches invalidate the cached window via
                   // `syncSessionConfigState`, which resets us back to the
                   // default so this branch runs again for the new model.
-                  if (session.contextWindowSize === DEFAULT_CONTEXT_WINDOW) {
+                  if (session.contextWindowSizeIsDefault) {
                     const inferred = inferContextWindowFromModel(model);
                     if (inferred !== null) {
                       session.contextWindowSize = inferred;
+                      session.contextWindowSizeIsDefault = false;
                     }
                   }
                 }
@@ -1740,14 +1746,16 @@ export class ClaudeAcpAgent {
               const nextUsage = totalTokens(lastAssistantUsage);
               if (nextUsage !== lastAssistantTotalUsage) {
                 lastAssistantTotalUsage = nextUsage;
-                await this.client.sessionUpdate({
-                  sessionId: params.sessionId,
-                  update: {
-                    sessionUpdate: "usage_update",
-                    used: nextUsage,
-                    size: session.contextWindowSize,
-                  },
-                });
+                if (!session.contextWindowSizeIsDefault) {
+                  await this.client.sessionUpdate({
+                    sessionId: params.sessionId,
+                    update: {
+                      sessionUpdate: "usage_update",
+                      used: nextUsage,
+                      size: session.contextWindowSize,
+                    },
+                  });
+                }
               }
             }
             for (const notification of streamEventToAcpNotifications(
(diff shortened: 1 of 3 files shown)


benbrandt:
Would this fix your issue? works much more reliably for me: https://github.com/agentclientprotocol/claude-agent-acp/pull/799

jsgrrchg:
Hi Ben, I tried it and it fixed the issue! Sorry if my code wasn't good, I tried my best, still learning :)

benbrandt:
all good! your code wasn't _wrong_ it just meant we'd have to wait longer to get the first one, which I tried to avoid.
My scenario still _might_ be wrong for certain providers, but I think it will be better in the general cases.

Thanks for flagging!

jsgrrchg:
Thanks for the feedback! You're right, your approach is much more elegant; mine was a bit too defensive 😅.

I tested it by switching from Opus to Sonnet in the middle of a conversation, and it still works correctly, so I think we're in the clear for the normal case.

Regarding the tradeoff, I don't have Bedrock, Vertex, or a corporate gateway setup to test against, but I'll keep an eye out for feedback from users on those configurations. For regular Claude Code subscriptions, it's working great. I use this ACP in here https://github.com/jsgrrchg/NeverWrite, and the experience has been amazing, I can even support elicitations! 

I can imagine it's quite hard to maintain an ACP implementation across all the clients it needs to support. You're doing excellent work, and I really admire it.

PR #820, opened 2026-06-29
fix: Emit tool_call before permission request

benbrandt:
The Claude Agent SDK changed the timing of `canUseTool` — it now runs *before* the assistant message's `tool_use` block streams to us. As a result we sent a permission request (and `AskUserQuestion` elicitation) referencing a `toolCallId` the client had not yet seen via a `tool_call` update, so clients couldn't associate the prompt with a tool call.

This supersedes #815, which fixed the ordering by blocking `canUseTool` until the stream emitted the `tool_call`. That approach could deadlock on cancellation (the `await` had no signal/timeout) and woke the waiter from only one branch, so suppressed tools never resolved it.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -280,6 +280,14 @@ type Session = {
    *  `tool_result` time so a long-running session doesn't accumulate every
    *  tool call for its whole lifetime. */
   toolUseCache: ToolUseCache;
+  /** Tracks which tool_use ids we've already emitted a `tool_call` for, so the
+   *  second source to encounter a tool call sends a `tool_call_update` instead
+   *  of a duplicate `tool_call`. The SDK can invoke `canUseTool` (→ a permission
+   *  request, which emits the tool_call eagerly so the client has it before
+   *  being asked to approve it) either before or after the assistant message's
+   *  tool_use block streams; this set makes the two paths converge regardless of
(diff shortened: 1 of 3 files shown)


benbrandt:
@SteffenDE does this work for you?

SteffenDE:
Well, I just tested this and it looks like the order is changed again starting with claude-agent-acp 0.50.0, so it is kind of solved for us. But if - for whatever reason - the SDK changes the order again, this PR solves half of the problem. The other half being that canUseTool does not include the `parentToolUseId` we use to associate a tool call with a subagent. So if the tool was not emitted yet, we'd send a tool_call, but it would not be associated with the correct `parentToolUseId`. Last time I checked the canUseTool parameters, that information wasn't there at all...

So this is fine for anyone not reading the _meta, but if ACP adds proper subagent support at some point, we could have the same issue again.

benbrandt:
ah they added this recently... fix incoming

benbrandt:
@SteffenDE nevermind they give us the subagent id not the parent :( 

So will this cause you _more_ issues if I merge? Should we leave as-is now that they fixed it?

SteffenDE:
I think it's fine to merge. Right now it doesn't make a difference and if it really changes again in the future, we might just kill subagent rendering as we only implemented it for Claude so far.

PR #853, opened 2026-07-08
feat: implement configurable LLM providers

Rizzen:
According to [custom endpoints rfd](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/docs/rfds/custom-llm-endpoint.mdx)
Implement the Configurable LLM Providers RFD so clients can discover and route the agent's LLM traffic through their own proxy or gateway without relying solely on the ad-hoc gateway auth _meta mechanism.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -9,10 +9,15 @@ import {
   CompleteElicitationNotification,
   CreateElicitationRequest,
   CreateElicitationResponse,
+  DisableProviderRequest,
+  DisableProviderResponse,
   ForkSessionRequest,
   ForkSessionResponse,
   InitializeRequest,
   InitializeResponse,
+  ListProvidersRequest,
+  ListProvidersResponse,
+  LlmProtocol,
   ListSessionsRequest,
   ListSessionsResponse,
   LoadSessionRequest,
@@ -25,9 +30,12 @@ import {
   PermissionOption,
   PromptRequest,
   PromptResponse,
+  ProviderInfo,
   ReadTextFileRequest,
   ReadTextFileResponse,
   RequestError,
+  SetProviderRequest,
+  SetProviderResponse,
   RequestPermissionRequest,
   RequestPermissionResponse,
   ResumeSessionRequest,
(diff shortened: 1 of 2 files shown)


src/acp-agent.ts
```diff
@@ -1028,11 +1088,112 @@ export class ClaudeAcpAgent {
...
+   * current non-secret routing. Headers are never echoed (they may hold
+   * secrets); only `apiType`/`baseUrl` are surfaced for UI display.
+   */
+  async unstable_listProviders(_params: ListProvidersRequest): Promise<ListProvidersResponse> {
+    const config = this.resolveProviderConfig();
+    const provider: ProviderInfo = {
+      providerId: PROVIDER_ID,
+      supported: SUPPORTED_PROTOCOLS,
+      required: true,
+      current: config ? { apiType: config.apiType, baseUrl: config.baseUrl } : null,
```
benbrandt, line 1105: > current: null` here means "unconfigured, using the agent's own default routing", but the schema defines it as disabled.**
>
> The schema doc on `ProviderInfo.current` says "Null or omitted means provider is disabled", and the RFD is stricter (behavior rule 5): a provider with `current: null` "is disabled and MUST NOT be used by the agent for LLM calls."
>
> So a fresh agent (normal Claude login, no `providers/set` or gateway auth) reports `{ required: true, current: null }` — per the spec that's a contradiction (a non-disableable provider that's disabled), and a strictly conformant client could render "main" as disabled or block prompting until a provider is configured, even though sessions work fine on default routing.
>
> Most clients will probably read `null` as "unconfigured" (which is what we mean), so nothing is broken today — but it'd be good to remove the ambiguity. Options:
> - raise it upstream as an RFD clarification — the spec currently has no state for "enabled but agent-managed", which any agent with built-in default routing will hit, or
> - report the actual effective default (`{ apiType: "anthropic", baseUrl: "https://api.anthropic.com" }`) as `current` — technically "current effective routing", though it risks implying client-managed routing is active when it isn't.


@anna239 got this from an agent review, I wonder if we need to handle this in the RFD somehow

AlexandrSuhinin: I rather don't understand why it has "required=true" if agent may be configured in other ways.
Disabling provider should be possible and will mean "unconfigured" -> default route.
Also I don't see reason why codex is treated differently then: https://github.com/agentclientprotocol/codex-acp/pull/272/changes#diff-ff482757b43f46e95ff8c07a2fca96aeab812e7d5d5f2076fbfedf81bdcfbd20R290

AlexandrSuhinin (changes requested):

src/acp-agent.ts
benbrandt: Yeah that would also solve the concern

AlexandrSuhinin (approved):

PR #872, opened 2026-07-13
feat: support rewinding to a message when forking a session

nitayart:
### What

`session/fork` now honors an optional `_meta.claudeCode.rewindTo` — the ACP message id (the `messageId` clients receive in `agent_message_chunk` updates and replayed history) of the assistant turn to fork from. The forked session's history ends at that turn; everything after it is not part of the fork.

Support is advertised as `agentCapabilities._meta.claudeCode.rewind: true` so clients can feature-detect.

### How

This wires up plumbing the codebase already prepared:

- `Session.messageIdToUuid` (previously documented as "NOT READ YET — recorded now so the mapping exists if/when we wire up fork/rewind") translates the client-visible ACP id back to the SDK message uuid.
- The uuid is passed to the Agent SDK as `resumeSessionAt` alongside the existing `forkSession: true` in the query options.
- For sessions whose history this process hasn't observed (e.g. forking straight from disk), a new `resolveMessageUuid` helper falls back to a `getSessionMessages` scan, mirroring `replaySessionHistory`'s mapping logic (last-write-wins keeps the turn-boundary uuid when one `msg_…` spans several content-block messages).
- Unknown ids fail with JSON-RPC invalid params.

No protocol changes: `_meta` is the sanctioned extension point, matching the existing `_meta.claudeCode.options` pattern on `session/new`.

### Why

Clients building conversation-branching UX (rewind to turn N, explore an alternative continuation) currently have no way to express it over ACP even though the underlying SDK supports it (`resumeSessionAt`). Fork-at-anchor is the minimal surface that unlocks it without touching the spec.

### Tests

- `src/tests/fork-rewind.test.ts` (5 cases): meta → `resumeSessionAt` translation via history scan; in-memory map preferred over history read; last-write-wins on split messages; invalid-params on unknown id; plain fork unchanged.
- Full suite: 576 passed | 20 skipped. `tsc` and eslint clean.
- Verified live against Claude Code 2.1.207: planted "favorite color is blue", overwrote with "red", forked with `rewindTo=<turn-1 message id>` — the original session answers "Red", the rewound fork answers "Blue".

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -398,8 +398,8 @@ type Session = {
    *  naturally yields the turn-boundary uuid when one `msg_…` spans several
    *  content-block messages.
    *
-   *  NOT READ YET — recorded now so the mapping exists if/when we wire up
-   *  fork/rewind. */
+   *  Read by `session/fork` when `_meta.claudeCode.rewindTo` is set — see
+   *  `resolveMessageUuid`. */
   messageIdToUuid: Map<string, string>;
 };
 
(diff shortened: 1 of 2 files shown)


benbrandt:
ALready handled I believe

PR #881, opened 2026-07-15
Add structured Bash titles and nested subagent transcripts

nikita-ashihmin:
## Motivation

ACP clients currently have two gaps when rendering Claude Code activity:

1. Bash tool calls expose the shell command as the standard ACP title, but the concise human-readable `description` produced by Claude Code is lost.
2. ACP 1.2 has no standard subagent tool kind or nested-message relationship, so subagent prose and child tools are either flattened into the main conversation or filtered out entirely.

This PR adds namespaced metadata and an explicit client capability so richer clients can render both cases without changing the standard ACP schema or regressing legacy clients.

### Bash descriptions

- Keeps the shell command in the standard ACP `title`, preserving command-oriented clients.
- Exposes the optional Claude Code Bash `description` as `_meta.claudeCode.title`.
- Applies the metadata consistently to eager permission tool calls, normal tool calls, and streamed input refinements.
- Falls back to the command title when no description is available.

### Nested subagent transcripts

- Adds the opt-in client capability `clientCapabilities._meta["subagent-transcript"] = true`.
- Enables the Claude Agent SDK `forwardSubagentText` option when the client advertises support.
- Preserves the pre-existing caller-provided `forwardSubagentText: true` option for compatibility.
- Marks Agent and legacy Task launches with `_meta.claudeCode.subagent = true`.
- Relates nested text, thinking, and child tool calls to the launching Agent/Task call through `_meta.claudeCode.parentToolUseId`.
- Preserves parent attribution during session-history replay, not only during live streaming.
- Handles nested Agent-to-Agent launches without inferring subagents from the generic ACP `think` kind.

## Compatibility

Clients that do not advertise `subagent-transcript` retain the previous behavior: subagent text and thinking remain filtered from the top-level feed.

The normal Agent/Task tool result is still emitted in every mode as the protocol-compatible fallback. No existing result or standard ACP field is removed or redefined.

## Validation

- `npm run build`
- `npm run check`
- `npm run test:run`
- 623 tests passed, 20 skipped

Coverage includes live and replayed nested text/thinking, child tool attribution, Agent/Task launch markers, nested Agent launches, legacy filtering, and SDK-option compatibility.

### DIFF
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -155,7 +155,7 @@ export function toolInfoFromToolUse(
     case "Bash": {
       const input = toolUse.input as BashInput | undefined;
       return {
-        title: input?.command ? input.command : "Terminal",
+        title: input?.description ?? input?.command ?? "Terminal",
         kind: "execute",
         content: supportsTerminalOutput
           ? [{ type: "terminal" as const, terminalId: toolUse.id }]
--- a/src/tools.ts
+++ b/src/tools.ts
@@ -155,7 +155,7 @@ export function toolInfoFromToolUse(
     case "Bash": {
       const input = toolUse.input as BashInput | undefined;
       return {
-        title: input?.description ?? input?.command ?? "Terminal",
+        title: input?.command ? input.command : "Terminal",
         kind: "execute",
         content: supportsTerminalOutput
           ? [{ type: "terminal" as const, terminalId: toolUse.id }]
--- a/README.md
+++ b/README.md
@@ -12,11 +12,23 @@ This tool implements an ACP agent by using the official [Claude Agent SDK](https
 - Following
 - Edit review
 - TODO lists
+- Nested subagent transcripts
 - Interactive (and background) terminals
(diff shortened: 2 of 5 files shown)


src/tools.ts
```diff
@@ -156,7 +156,7 @@ export function toolInfoFromToolUse(
     case "Bash": {
       const input = toolUse.input as BashInput | undefined;
       return {
-        title: input?.command ? input.command : "Terminal",
+        title: input?.description ?? input?.command ?? "Terminal",
```
Ololoshechkin: previously, in IDE the `title` field for "execute" tool was rendered as a shell command preview, hence it was expecting `command`. Has it changed?

nikita-ashihmin:
@Ololoshechkin Good point. I restored the existing ACP `title` behavior, so it remains the shell command preview. The Claude-provided description is now exposed separately as `_meta.claudeCode.title` for clients that want the human-readable heading.

Ololoshechkin (approved):

PR #892, opened 2026-07-19
feat: opt-in forwardSubagentText to surface subagent text/thinking

pandakeeper0:
By default the bridge drops subagent (parent_tool_use_id != null) text and thinking blocks to keep subagent prose out of the top-level feed. Add a per-session opt-in, read from the client-provided forwardSubagentText option, that forwards the full subagent content so consumers can render a nested transcript. Default false preserves upstream behavior.


benbrandt:
I believe there is already work in this direction happening in #881

PR #915, opened 2026-07-25
feat: Add _claude/usage extension method for structured rate-limit usage

Orcaryn:
Adds a `_claude/usage` ACP extension method that exposes the SDK's structured `/usage` data, so a client can render plan rate-limit utilization (5-hour, 7-day, …) without scraping the `/usage` slash-command text.

### Shape

A thin passthrough. The caller asks for the windows it wants and gets them back verbatim:

```jsonc
// → _claude/usage
{ "keys": ["five_hour", "seven_day"] }

// ← result
{
  "five_hour": { "utilization": 42, "resets_at": "2026-01-01T00:00:00Z" },
  "seven_day": { "utilization": 7,  "resets_at": "2026-01-07T00:00:00Z" }
}
```

Key selection and labelling stay entirely on the caller, so windows the SDK adds later need no change here.

- Windows the SDK does not report are omitted from the result.
- The result is `{}` when plan limits do not apply — API key, Bedrock, Vertex (`rate_limits: null`).
- An unfed `Pushable` holds an ephemeral query open just long enough to issue the `get_usage` request; it is closed in a `finally`, so a failed request leaks no subprocess.

Params are validated by `parseUsageRequest`, mirroring the existing `parseSteerRequest` — non-object params, or a `keys` that isn't an array of strings, raise `invalidParams`.

### Notes for review

The SDK method behind this is `usage_EXPERIMENTAL_MAY_CHANGE_DO_NOT_RELY_ON_THIS_API_YET()`, hence an extension method rather than a core one. Two things I'm happy to change on request:

- **Naming** — `_claude/usage` mirrors codex-acp's `_codex/account/rate_limits`.
- **Discovery** — unlike `_session/steering`, this advertises no capability flag. I can gate it behind an `InitializeResponse._meta` entry if you'd prefer clients feature-detect it.

### Testing

`src/tests/usage.test.ts` covers key selection, unreported windows, the no-plan-limits case, and query cleanup when the usage request throws. Full CI run locally — `format:check`, `lint`, `build`, `test:run` (656 passed, 20 skipped) — all green.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -199,6 +199,9 @@ const TURN_NO_RESULT_MESSAGE =
  *  `InitializeResponse._meta.steering.supported`. */
 const STEER_METHOD = "_session/steering";
 
+/** Extension method returning the caller-requested claude.ai rate-limit windows. */
+const USAGE_METHOD = "_claude/usage";
+
 /** How urgently the SDK delivers a steered message relative to the running
  *  turn — an internal Claude implementation detail, not part of the wire
  *  contract. `now` pre-empts the current generation and handles the message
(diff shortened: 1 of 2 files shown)


Orcaryn:
Withdrawing this — turns out the stock adapter already covers our use case. `/usage` is a supported command that returns its report as a normal agent message, so a client can read the plan windows without any new method. Closing rather than leaving it open for review. Thanks!

PR #919, opened 2026-07-26
fix(steering): add opt-in host-owned fallback

### LINKED ISSUE #903: Steering `startedNewTurn` detaches `prompt()` without an observable terminal response
## Summary

When `_session/steering` reaches a session after the original turn has already
settled, `claude-agent-acp` starts a new turn through a detached `prompt()` call
and returns:

```json
{ "outcome": "startedNewTurn" }
```

The new turn executes and emits `session/update` notifications, but no external
`session/prompt` request owns its lifecycle. Its successful `PromptResponse` is
discarded, while failures are only logged by the adapter.

Hosts that scope streaming, cancellation, persistence, and UI state to a pending
`session/prompt` cannot reliably consume or terminate this detached turn.

## Environment

- `@agentclientprotocol/claude-agent-acp`: `0.61.0`
- `@anthropic-ai/claude-agent-sdk`: `0.3.217`
- Extension method: `_session/steering`

## Reproduction

1. Create an ACP session.
2. Start a normal `session/prompt`.
3. Send `_session/steering` near the original turn's completion boundary.
4. Let the original turn settle before the steering request is handled.
5. Observe that steering returns `startedNewTurn`.
6. Observe that Claude starts another API request and produces output.
7. Observe that no ACP request receives the new turn's terminal response.

In one observed run:

- `13:08:42`: steering triggered the follow-up Claude API request.
- `13:08:50`: the first response content arrived.
- The content was not delivered to the client-owned run because the original
  `session/prompt` lifecycle had already completed.

## Root cause

The idle branch starts `prompt()` without returning or awaiting its result:

```ts
this.prompt(promptRequest).catch((error) => {
  this.logger.error(`Session ${sessionId}: steered new turn failed: ${error}`);
});
return { outcome: "startedNewTurn" };
```

Although the turn uses the adapter's internal `prompt()` machinery, there is no
external JSON-RPC request waiting for its `PromptResponse`.

## Expected behavior

Steering should inject only into an existing in-flight turn.

If no turn is running, the adapter should not start a detached turn. It should
return an explicit result such as:

```json
{ "outcome": "promptRequired" }
```

`promptRequired` must guarantee that the adapter did not enqueue, inject, or
otherwise consume the supplied content. The Host can then submit the same
content through a normal `session/prompt` request using the same `sessionId`.

This preserves the standard lifecycle:

```text
session/prompt
  -> session/update*
  -> PromptResponse or JSON-RPC error
```

Using the same session preserves the existing conversation context.

A typed `no_running_turn` error would also be valid if its contract clearly
states that the content was not consumed and should be retried through
`session/prompt`.

## Proposed fix

1. Keep the active-turn path unchanged:
   - Push the steering message to the existing SDK input.
   - Preserve `priority: "now"`.
   - Return `injected`.
   - Keep output and completion attached to the original `session/prompt`.

2. Change the idle path:
   - Do not call `this.prompt()`.
   - Do not push content into the SDK input.
   - Return `promptRequired` or a typed `no_running_turn` error.

3. Advertise the idle behavior through the steering capability so Hosts can
   negotiate the updated result contract.

4. Update `examples/steering.ts` and relevant public documentation.

## Test coverage

Add tests proving that:

- Steering an active turn still returns `injected`.
- Active steering still uses `priority: "now"`.
- Steering an idle session does not call `prompt()`.
- The idle path does not push content into the SDK input.
- The idle result clearly states that the content was not consumed.
- The same content can subsequently be sent through `session/prompt`.
- The normal prompt receives its updates and terminal response.
- Query-closed sessions preserve the existing session-ended error.
- No detached prompt remains after the idle steering path.

## Out of scope

This issue does not propose:

- A general ACP turn-completion notification.
- New `turn_started` or `turn_completed` protocol events.
- Changes to background-task lifecycle handling.
- Client or adapter idempotency keys.
- Changes to Claude SDK session state semantics.

## Related implementation

- #885 introduced the current steering implementation.
- #871 is the original steering feature request.

This issue is limited to the lifecycle gap in the `startedNewTurn` fallback.

ran411285752:
## Summary

Add an opt-in host-owned fallback for the `_session/steering` race while preserving the existing default contract.

When a Host explicitly requests `promptRequired` in the steering request metadata, the Adapter leaves the content unconsumed so the Host can submit it through a normal `session/prompt`. Hosts that do not opt in continue to receive the existing `startedNewTurn` behavior.

## Problem

The idle fallback calls `prompt()` asynchronously and returns `startedNewTurn` immediately.

Although that new turn executes and emits `session/update` notifications, no external `session/prompt` request owns its lifecycle. Its terminal `PromptResponse` is discarded, and failures are only written to the Adapter log.

Hosts that scope streaming, cancellation, persistence, and UI state to a pending prompt request cannot reliably consume or terminate the detached turn.

## Root cause

`startedNewTurn` creates a new prompt lifecycle inside an already-completed steering request:

```ts
this.prompt(promptRequest).catch(...);
return { outcome: "startedNewTurn" };
```

ACP already has a complete lifecycle for new turns:

```text
session/prompt
  -> session/update*
  -> PromptResponse or JSON-RPC error
```

The detached call bypasses the Host side of that lifecycle. `session/update` is session-scoped and cannot replace the missing request-owned terminal response.

## Design rationale

Steering is an operation on an existing in-flight turn, not an alternative transport for starting a new turn.

When a running turn exists, the Adapter injects the message into that turn with `priority: "now"` and returns `injected`.

When no running turn exists, the Host-owned behavior is safer because the Host can explicitly create a standard `session/prompt` lifecycle. However, the existing `startedNewTurn` result is already relied upon by Hosts, so the safer behavior is opt-in rather than a breaking default change.

Using the same `sessionId` preserves Claude's existing conversation context while restoring a standard request lifecycle for streaming, completion, errors, cancellation, usage, persistence, and UI state.

### Anthropic client architecture precedent

This ownership model is also consistent with the architecture used by Anthropic's shipped clients:

- Claude Desktop's local-session bridge starts a session and returns a `sessionId`; subsequent message delivery, interruption, queued-message operations, and events remain explicitly session-scoped.
- The official Claude Code VS Code extension creates a client-addressable channel before starting its query. Input, streamed output, interruption, and teardown are routed through that channel. An unknown or closed channel does not silently create a replacement query.

These clients do not implement this ACP extension, so this is non-normative architecture precedent rather than a protocol compatibility claim. The relevant property is that the client establishes and owns the lifecycle before work begins.

## Fix

- Preserve the active-turn behavior:
  - push the steering message into the existing SDK input;
  - preserve `priority: "now"`;
  - return `injected`;
  - keep completion attached to the original `session/prompt`.

- Preserve the legacy idle default:
  - without opt-in, keep the detached `prompt()` fallback;
  - return `startedNewTurn`.

- Add request-level opt-in:

```json
{
  "sessionId": "...",
  "prompt": [...],
  "_meta": {
    "steering": {
      "idleBehavior": "promptRequired"
    }
  }
}
```

- For the opt-in path only:
  - do not call `prompt()`;
  - do not push SDK input;
  - do not mutate `turnQueue`;
  - return `promptRequired` with reason `noRunningTurn`.

- Keep the existing initialize capability declaration:

```json
{
  "_meta": {
    "steering": {
      "supported": true
    }
  }
}
```

- Update the steering example and tests to cover both the legacy default and the host-owned opt-in path.

## Compatibility

This keeps the existing default wire contract for the private, pre-standardized 0.x steering extension. Existing Hosts continue to receive `startedNewTurn` when they do not request another behavior.

Hosts that want the request-owned lifecycle can opt into `promptRequired` through request `_meta` and then resubmit the unchanged content through `session/prompt`.

Standard ACP methods are unchanged, and active-turn steering still returns `injected`.

## Verification

- Focused steering tests: 8 passed.
- The new default-compatibility tests failed before the implementation change and pass with the opt-in implementation.
- `npm run lint`: passed.
- `npm run build`: passed.
- Standalone example TypeScript check with Node types: passed.
- Full Windows suite: 648 passed, 6 failed.
- The same six path-separator assertions reproduce on clean `origin/main` (646 passed, 6 failed); Linux CI remains the final full-suite check.

## Out of scope

- A general ACP turn lifecycle or new `turn_started` / `turn_completed` events.
- Background-task lifecycle changes.
- Claude SDK idle-state semantics.
- Steering idempotency keys.
- Changes to standard ACP methods.

Fixes #903

### DIFF
--- a/examples/steering.ts
+++ b/examples/steering.ts
@@ -11,17 +11,15 @@
  * The wire protocol has three moving parts:
  *
  *   1. The agent advertises support in its `initialize` response, at the
- *      top-level `_meta.steering` (a sibling of `agentCapabilities`), including
- *      `idleBehavior: "promptRequired"` so the client knows the idle contract.
+ *      top-level `_meta.steering` (a sibling of `agentCapabilities`).
  *   2. The client calls the `_session/steering` request with `{ sessionId,
  *      prompt }` while a turn is running.
  *   3. The agent replies with an `outcome`:
  *        - "injected"       the message joined the running turn;
- *        - "promptRequired" the turn had already finished (an unavoidable race),
- *                           so the message was NOT consumed and the client must
- *                           start a normal `session/prompt` to send it.
- *      Both are success outcomes: the Adapter either consumes the message in
- *      the running turn or leaves it Host-owned for explicit prompt delivery.
+ *        - "startedNewTurn" the turn had already finished and the legacy
+ *                           detached fallback was used;
+ *        - "promptRequired" the turn had already finished, but this request
+ *                           explicitly opted into Host-owned prompt delivery.
  *
  * This example launches the agent as a subprocess, starts a deliberately
  * long-running prompt, and — as soon as the agent begins streaming — injects a
@@ -58,6 +56,7 @@ const STEERING_METHOD = "_session/steering";
 type SteeringRequest = {
   sessionId: string;
   prompt: Array<{ type: "text"; text: string }>;
+  _meta?: { steering?: { idleBehavior?: "promptRequired" } };
 };
 
 /** Result of a `_session/steering` request. `injected` means the message joined
@@ -65,14 +64,14 @@ type SteeringRequest = {
  *  message was NOT consumed and must be (re)sent through a normal
  *  `session/prompt`. Both are successes. */
 type SteeringResponse =
-  { outcome: "injected" } | { outcome: "promptRequired"; reason: "noRunningTurn" };
+  | { outcome: "injected" }
+  | { outcome: "startedNewTurn" }
+  | { outcome: "promptRequired"; reason: "noRunningTurn" };
 
-/** The steering capability advertised at the top-level `_meta.steering` of the
- *  `initialize` result. `idleBehavior` narrows the idle contract this example
- *  relies on. */
+/** The existing steering capability advertised at the top-level `_meta.steering`
+ *  of the `initialize` result. The idle behavior is selected per request. */
 type SteeringCapability = {
   supported?: boolean;
-  idleBehavior?: "promptRequired";
 };
 
 // The built agent entry. Run `npm run build` first so this exists.
@@ -148,10 +147,10 @@ async function main() {
       });
       const steering = (init._meta as { steering?: SteeringCapability } | null | undefined)
         ?.steering;
-      if (steering?.supported !== true || steering.idleBehavior !== "promptRequired") {
-        throw new Error("agent does not advertise the promptRequired steering contract");
+      if (steering?.supported !== true) {
+        throw new Error("agent does not advertise steering support");
       }
-      log(`agent steering idle behavior: ${steering.idleBehavior}`);
+      log("agent advertises steering support");
 
       // 2. Open a session.
       const { sessionId } = await agent.request(methods.agent.session.new, {
@@ -180,6 +179,9 @@ async function main() {
       const steerRequest: SteeringRequest = {
         sessionId,
         prompt: [{ type: "text", text: STEER }],
+        // Opt into the host-owned idle fallback. Without this request metadata,
+        // the Adapter preserves its legacy `startedNewTurn` behavior.
+        _meta: { steering: { idleBehavior: "promptRequired" } },
       };
       const result = await agent.request<SteeringResponse>(STEERING_METHOD, steerRequest);
       log(`steer outcome: ${result.outcome}`);
--- a/examples/steering.ts
+++ b/examples/steering.ts
@@ -20,8 +20,8 @@
  *        - "promptRequired" the turn had already finished (an unavoidable race),
  *                           so the message was NOT consumed and the client must
  *                           start a normal `session/prompt` to send it.
- *      Both are success outcomes — the message is never dropped and the race is
- *      never surfaced as an error.
+ *      Both are success outcomes: the Adapter either consumes the message in
+ *      the running turn or leaves it Host-owned for explicit prompt delivery.
  *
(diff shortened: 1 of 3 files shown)


Rizzen:
Thanks for working on this and for the thoughtful contribution.

I’d prefer not to change the current default behavior, since existing hosts are already relying on `startedNewTurn`.

Could we instead introduce an optional steering mode in `_meta` that opts into the rejection behavior described here? For example, a host could explicitly request something like:
```
{
  "_meta": {
    "steering": {
      "idleBehavior": "promptRequired"
    }
  }
}
```

When this mode is enabled and there is no running turn, the adapter would return `promptRequired` without consuming the content, allowing the host to resubmit it through `session/prompt`. When the flag is absent, the adapter would preserve the existing behavior for backward compatibility.

This would provide the safer host-owned lifecycle without introducing a breaking default change. What do you think about this approach?

Rizzen (changes requested):
Please check my message above, we cannot break current contract on steering

ran411285752:
Thanks, that makes sense. We agree that the existing default contract should remain unchanged for compatibility.

We will revise this PR to:

- keep the existing `startedNewTurn` behavior when no steering mode is requested;
- accept request-level `_meta.steering.idleBehavior: "promptRequired"` as an explicit opt-in;
- return `promptRequired` without consuming the content only for Hosts that opt in;
- keep the existing `supported: true` capability declaration without adding new capability fields;
- update the tests and example to cover both the legacy default and the host-owned opt-in path.

We will push the compatibility update to this PR.

Rizzen (approved):

Rizzen:
Thank you!

Issue #825, opened 2026-06-30
Prompt loop never terminates a turn when the model stream ends abnormally; session hangs 'running' until the next prompt drains it (0.52.0, sdk 0.3.191)


Seluj78:
## Summary

The adapter's prompt loop does not terminate a turn when the underlying model stream ends abnormally (connection dropped / closed mid-response, or an error that arrives without a clean `result`). When this happens:

- the notification stream stops mid-turn (last `session/update` is a partial `agent_message_chunk`),
- no `PromptResponse` (`result`) is ever returned for the in-flight `session/prompt`,
- the SDK `query` and the adapter both go fully idle,
- the session sits "running" forever from the client's perspective.

The work is recoverable only by sending a *second* `session/prompt`: the next `query.next()` consumes the stale terminal state left by the dead turn, the prior turn is silently abandoned, and the new prompt runs. This is the same defect family as #435, #654, and #688, root-caused in (unmerged) #659 / #660, and it still reproduces on the current release.

## Impact

High. A single transient model-stream drop mid-turn wedges the session indefinitely with no error surfaced to the client. There is no timeout and no terminal event, so a client that waits on the `session/prompt` result hangs until a human notices and pokes the session with another prompt. Output produced just before the drop is lost (never flushed, never finalized).

## Environment

- `@agentclientprotocol/claude-agent-acp` `0.52.0` (also present in `0.53.0`; its changelog only adds ACP logout, an SDK bump, and tool_call/permission ordering, none of which touch the prompt loop)
- bundled `@anthropic-ai/claude-agent-sdk` `0.3.191`
- node `v22.18.0`, macOS 26.5.1 (arm64)

## Evidence (live process forensics)

A real session wedged for ~40 minutes. Captured before recovering it.

Notification log (client side) stops mid-turn with no terminator:

```
seq 1064  Stopped: prompt_complete        ← a prior turn, healthy
...                                         (later turn runs, tool calls, chunks)
seq 1236  AgentMessageChunk "Real, user-visible one. Let me read"
seq 1237  AgentMessageChunk "both sides: ... reference + ..."   ← cut mid-sentence
          ← nothing further. No result. No Stopped. Session "running".
```

Process tree while wedged:

```
aoe runner ── node claude-agent-acp (adapter, pid 48980) ── claude SDK (pid 48981) ── MCP servers
```

Both node processes are parked in the event loop, **not** blocked on any read/write syscall:

[...]

PR #937, opened 2026-08-01
fix(steering): retain prompt ownership after injection

### LINKED ISSUE #934: Injected steering can outlive its owning session/prompt
## Summary

With `@agentclientprotocol/claude-agent-acp` 0.64.0 and `@anthropic-ai/claude-agent-sdk` 0.3.220, an active `_session/steering` request can return `injected` and successfully change the model output, while the original `session/prompt` rejects before the steered response starts streaming.

The later response continues through `session/update`, but no pending prompt request remains to own its terminal response.

This is the active `injected` path, not the idle `promptRequired` fallback from #903.

## Environment

- OS: Windows
- Node.js: 22.16.0
- `@agentclientprotocol/claude-agent-acp`: 0.64.0 (`v0.64.0`, clean checkout)
- `@anthropic-ai/claude-agent-sdk`: 0.3.220
- `@agentclientprotocol/sdk`: 1.3.0
- Provider: Anthropic-compatible endpoint configured through the normal Claude settings environment

A baseline request without steering completed normally with `stopReason: end_turn`.

## Reproduction

1. Build a clean `v0.64.0` checkout with `npm ci && npm run build`.
2. Initialize the agent and create a session.
3. Start a long-running `session/prompt`, but do not await it yet.
4. While that prompt is active, send `_session/steering` on the same session:

```json
{
  "sessionId": "<same session>",
  "prompt": [
    {
      "type": "text",
      "text": "Stop the previous task and reply with exactly: STEERED-REAL-OK"
    }
  ],
  "_meta": {
    "steering": {
      "idleBehavior": "promptRequired"
    }
  }
}
```

5. Continue listening for `session/update` notifications even if the original `session/prompt` settles.

The repository's `examples/steering.ts` can be used for steps 2-4. To observe step 5, catch the original prompt rejection and keep the ACP connection open briefly instead of exiting immediately.

## Actual result

Observed real-request ordering (UTC):

```text
05:15:17.318  steering request sent
05:15:17.320  steering response: { outcome: "injected" }
05:15:17.327  original session/prompt rejected:
              Internal error: [ede_diagnostic] result_type=user last_content_type=n/a stop_reason=null

05:15:24.153  steered agent_thought_chunk updates begin
05:15:25.320  steered agent_message_chunk updates produce: STEERED-REAL-OK
05:15:25.364  steered usage_update arrives
05:15:25.371  session_info_update arrives
```

The connection remained open for another 60 seconds. No prompt-owned terminal response arrived for the steered output.

The active run reproduced the same immediate `injected` followed by the original prompt rejection in two attempts. The second attempt kept the connection open and captured the complete later response shown above.

## Expected result

When `_session/steering` returns `injected`, the running prompt lifecycle should not settle before the steered response. The owning `session/prompt` should remain observable through the steered output and settle with its terminal result, error, cancellation, and usage.

## Impact

Hosts that bind streaming, persistence, cancellation, usage, and UI completion to a pending `session/prompt` can close the turn before the steered response arrives. The model follows the steering instruction, but the later output has no request-owned terminal lifecycle.

ran411285752:
Fixes #934

## Problem

Active steering (`_session/steering` with outcome `injected`) previously let the interrupted command's terminal result settle (or reject) the original ACP v1 `session/prompt` before the injected response even started. The steered output then streamed with no owning turn, so its own terminal result leaked a false trailing-idle debt and the client observed the original prompt resolving early (even with zero usage) or rejecting.

## Scope

This change is limited to ACP v1. The existing `_session/steering` wire contract remains unchanged:

- The request still supports the optional `_meta.steering.idleBehavior: "promptRequired"` field.
- The response outcomes remain `injected`, `startedNewTurn`, and `promptRequired`.
- No `activeBehavior` field or new capability is introduced.

## Why `injected` keeps the original prompt pending

An `injected` outcome means the steering message joined the current prompt turn, not a new one. The SDK cancels the running generation to process the `priority: "now"` message, emits an interrupted boundary result, then replays the injected user echo and runs the steered generation. The original `session/prompt` is the request that owns all of that: it stays pending through the interrupted result and the injected echo, and only the steered terminal result settles it.

## The exact SDK delivery path (unchanged)

- `userMessage.uuid = randomUUID()` — the injected `SDKUserMessage` keeps its freshly generated UUID;
- `userMessage.priority = "now"` (existing `STEER_PRIORITY`);
- `session.input.push(userMessage)` — the same input stream and conversation context as every other message in the turn.

No change to UUID generation, priority, the input stream, or context.

## Both premature-settlement variants fixed

1. **Early resolve** — the interrupted result (`stop_reason: null`) previously settled the prompt before the steered response. It is now classified as a superseded boundary result: its usage is still accumulated into the owning turn, but it never settles the prompt, never counts idle debt, and exits before any terminal handling.
2. **Diagnostic reject** — the same superseded classification also prevents the interrupted result's `[ede_diagnostic]` reject payload from failing the prompt.
3. **Held turn** — a turn already held open behind a live subagent (`Turn.deferredSettle`) previously got handoff-settled by `ensureActiveTurn` on the injected echo's next user-turn result, resolving the original prompt before the steered terminal result. `ensureActiveTurn` now keeps a retained-steering held turn held while the injection is in flight, and the steered terminal result clears the marker before `settleOrDefer`, so a subsequent still-live subagent followup treats it as an ordinary held turn again.

## Additional retained-turn lifecycle handling

4. **Held turn awaiting its echo goes unowed-idle** (High) — if the injected echo never replays, an unowed idle (`phase: "awaiting_echo"` with `owedTrailingIdles === 0`) on a held turn previously left the prompt pending forever behind a still-live subagent, or — once the subagent drained — settled it with the PRE-steering deferred outcome (wrong usage, no steered content). The idle handler now fails it with the standard `no_result` error before the ordinary held-turn branch, covering both live and drained cases (same contract as the #825 idle-fail).
5. **Steered refusal leaves a stale active marker** (Medium) — the `retainedSteering` marker was only cleared at the bottom of the result handler's normal path, so a steered refusal (which re-holds the turn behind a still-live subagent) left a stale `phase: "active"` marker that made `ensureActiveTurn` skip the next echo-less result's hand-off. The result handler now claims the active marker right after the superseded-result check and before all terminal early-return lanes; a result still awaiting its echo keeps the marker so an autonomous subagent followup landing before the echo stays part of the injected boundary.
6. **Autonomous followup settles a held prompt before steering completes** (High) — a drained subagent's autonomous followup result reached `settleDeferredIfDrained()` while the retained marker was still set, and the active marker was cleared even by an autonomous result. Either path resolved the held prompt with the PRE-steering deferred outcome (wrong usage, no steered content) before the injected echo / steered terminal result, so the steered output streamed outside the ended prompt. Three gates now hold while any `retainedSteering` state exists: the `phase: "active"` marker is claimed only by a user-driven result; the autonomous branch skips `settleDeferredIfDrained()`; and the held-turn idle branch skips it too (the autonomous followup's own trailing idle can't settle the prompt). Autonomous usage, `usage_update`, and idle debt are unchanged.
7. **Post-echo missing steered result leaves the prompt pending forever** (High) — the held no-result idle branch only matched `phase: "awaiting_echo"`. Once the injected echo replayed (`phase: "active"`), a missing steered terminal result with an unowed idle fell into the ordinary held branch, which skips `settleDeferredIfDrained()` while a retained marker exists and never fails the prompt — permanent `pending`. The no-result condition is now any `retainedSteering` state with `owedTrailingIdles === 0`: an unowed idle while the marker is set means the SDK turned over without the retained echo or the steered terminal result it owes, so it fails with the standard `no_result` error. An idle WITH debt still belongs to a counted result (an autonomous followup's trailer) and is absorbed by the ordinary held branch — the autonomous-result regression tests exercise exactly that.

## Verified result / echo / final-result / idle order

```
session_state_changed: running
user echo (original prompt)
result (stop_reason null)  -> superseded boundary, prompt still pending
user echo (injected UUID)  -> retained steering phase becomes active
steered assistant output   -> e.g. STEERED-REAL-OK
result (end_turn)          -> settles the original prompt with summed usage
session_state_changed: idle -> single trailing idle, absorbed by owed debt
```

For a held turn (deferred subagent), the same order holds with the prompt staying held through the steered result; if the subagent is still live at that point the turn re-defers and only settles when the subagent drains, at which point the summed usage of interrupted + steered commands is attributed to the original prompt.

## Unchanged idle behavior

The idle `startedNewTurn` and `promptRequired` behaviors are untouched. This fix changes only the active `injected` path, which now retains prompt ownership by default.

## Automated verification

- RED: the new #934 tests fail on the base commit (premature resolve, diagnostic reject, pre-activation, held-turn settle, missing echo, stale marker, autonomous followup before/after echo, post-echo missing result) and pass after the fix.
- Focused: `npx vitest run src/tests/acp-agent.test.ts -t "turn steering|session/cancel wedge recovery|session_state_changed|deferred settlement for live background subagents|autonomous followup lands|never emits its final result"` -> 51 passed; the whole file -> 324 passed, 6 failed (pre-existing path-separator), 9 skipped.
- `npm run build`, `npm run lint`, `git diff --check`, and `prettier --check` on the committed content of the changed files -> pass.
- `npm run test:run` -> 695 passed, 6 failed; the remaining failures are the pre-existing Windows path-separator assertions in `acp-agent.test.ts` and do not touch the Steering implementation.
- `npm run format:check` -> fails only on files that are already failing on `origin/main` (repo-wide prettier drift on this Windows host); the changed files are clean on their committed content.

## Real-model verification

The latest run on the final commit used the issue's prompts and showed the full order above: `injected` outcome, interrupted result, prompt still pending, injected echo, `STEERED-REAL-OK` output, terminal result, one final idle, and the original `session/prompt` resolving `end_turn` with the steered command's usage attributed to it (`input=93`, `output=41`, `cache_read=27904`, `total=28038`) — versus the pre-fix zero-usage early resolve. Two earlier runs showed the same ordering and ownership behavior.

### DIFF
--- a/examples/steering.ts
+++ b/examples/steering.ts
@@ -15,7 +15,15 @@
  *   2. The client calls the `_session/steering` request with `{ sessionId,
  *      prompt }` while a turn is running.
  *   3. The agent replies with an `outcome`:
- *        - "injected"       the message joined the running turn;
+ *        - "injected"       the message joined the running turn. The ORIGINAL
+ *                           `session/prompt` request stays pending while the
+ *                           SDK interrupts the running generation to handle the
+ *                           injected message; that interrupt's boundary result
+ *                           is folded into the same turn, and only the steered
+ *                           terminal result settles the original prompt — which
+ *                           is the request whose `PromptResponse` the client
+ *                           awaits. Its `usage` is the sum of the interrupted
+ *                           and steered commands;
  *        - "startedNewTurn" the turn had already finished and the legacy
  *                           detached fallback was used;
  *        - "promptRequired" the turn had already finished, but this request
@@ -60,9 +68,11 @@ type SteeringRequest = {
 };
 
 /** Result of a `_session/steering` request. `injected` means the message joined
- *  the running turn; `promptRequired` means the turn had already settled, so the
- *  message was NOT consumed and must be (re)sent through a normal
- *  `session/prompt`. Both are successes. */
+ *  the running turn — the original `session/prompt` remains pending through the
+ *  SDK's interrupted boundary result and the steered final result, and is the
+ *  request whose `PromptResponse` carries the turn's outcome. `promptRequired`
+ *  means the turn had already settled, so the message was NOT consumed and must
+ *  be (re)sent through a normal `session/prompt`. Both are successes. */
 type SteeringResponse =
   | { outcome: "injected" }
   | { outcome: "startedNewTurn" }
@@ -202,8 +212,12 @@ async function main() {
         log(`continuation stopReason: ${continuation.stopReason}`);
       }
 
-      // 5. Await the original turn. With outcome "injected" the steer already
-      //    reshaped the output above; the promptRequired branch owns its own turn.
+      // 5. Await the ORIGINAL turn — the pending session/prompt from step 3. On
+      //    "injected" this is the request that owns the steered outcome: it stays
+      //    pending through the interrupted boundary result and the steered
+      //    terminal result, and settles here with the combined usage. On
+      //    "promptRequired" the original turn settled on its own earlier and the
+      //    continuation branch above owns its separate turn.
       const response = await turn;
       log(`original turn stopReason: ${response.stopReason}`);
       process.stdout.write("\n----- end of agent output -----\n");
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -283,6 +283,14 @@ type Turn = {
   /** uuid stamped on the pushed `SDKUserMessage`; the SDK echoes it back so the
    *  consumer can match the replayed user message to this turn. */
   promptUuid: string;
+  /** Active Steering remains part of this ACP v1 prompt turn until the SDK
+   *  echoes the latest injected UUID and then emits its terminal result. The
+   *  newest accepted steering message overwrites the previous one, so only the
+   *  latest UUID can become the terminal boundary (see steer()). */
+  retainedSteering?: {
+    promptUuid: string;
+    phase: "awaiting_echo" | "active";
+  };
   /** Local-only slash commands (e.g. `/clear`) return a result without an echo,
    *  so the consumer can't promote them via the replay; it falls back to
    *  promoting the queue head when the result arrives. */
@@ -1904,8 +1912,8 @@ export class ClaudeAcpAgent {
     // which is exactly the window in which steering is meaningful. This check
     // and the active-path push below stay in one synchronous section so the
     // turn cannot settle in the gap between deciding to inject and enqueueing.
-    const turnInFlight = (session.turnQueue ?? []).some((turn) => !turn.settled);
-    if (!turnInFlight) {
+    const runningTurn = (session.turnQueue ?? []).find((turn) => !turn.settled);
+    if (!runningTurn) {
       const promptRequest: PromptRequest = {
         sessionId,
         prompt: params.prompt,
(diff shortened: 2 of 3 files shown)


ran411285752:
Closing this PR because #958 has since landed and supersedes its core steering state machine.

#958 now owns injected steering settlement through `steeredEchoes` / `steeredSettle` and settles the turn at SDK `idle`. Keeping this PR's separate `retainedSteering` state machine would duplicate ownership and create conflicting settlement paths.

I verified the #937 regression cases against current `main`: 10 pass and 7 fail. Most failures reflect the intentionally different missing-echo semantics adopted by #958 or tests coupled to the old private state. Two focused correctness gaps remain and will be addressed in a small follow-up based directly on #958:

- an interrupted diagnostic result before the injected echo can incorrectly reject the owning prompt;
- after the injected echo, an `idle` without a steered final result can incorrectly reuse the pre-echo result and report success.

The follow-up will preserve #958's state machine and add only the minimal guards and regression tests for those two cases. Thanks to the maintainers for landing the broader fix in #958.

ran411285752:
The focused follow-up is now available as #969.

PR #949, opened 2026-08-03
fix(adapter): dedupe local-only command result output

### LINKED ISSUE #948: Local-only slash command output is emitted twice via local_command_output and result
## Description

For local-only slash commands such as `/context`, the Claude Agent SDK can expose the same command output through both:

1. a `system` message with subtype `local_command_output`; and
2. the terminal `result` message for the turn.

The adapter currently forwards both values as ACP `agent_message_chunk` notifications. Because both notifications have the same content and no remaining source distinction, ACP clients render the complete command output twice.

This was observed with `claude-agent-acp` 0.63.0. The two forwarding paths are also still present on current `main` (`08e0e7f`, package version 0.64.2).

## Steps to reproduce

1. Start `claude-agent-acp` through an ACP client that records `session/update` notifications.
2. Send `/context` through `session/prompt`.
3. Inspect the emitted `agent_message_chunk` notifications.

When the SDK emits both forms, the observable ACP sequence is equivalent to:

```text
SDK system/local_command_output("<full command output>")
  -> ACP agent_message_chunk("<full command output>")

SDK result/success(result = "<full command output>")
  -> ACP agent_message_chunk("<full command output>")
```

## Actual behavior

The ACP client receives and renders two identical copies of the complete local-command output.

## Expected behavior

The ACP client should receive the local-command output exactly once.

The adapter must still emit the output when the SDK provides only one of the two forms. In particular, the terminal `result` path remains necessary for SDK/version/provider combinations that do not emit `local_command_output`.

## Root cause

The `local_command_output` handler forwards its content directly:

https://github.com/agentclientprotocol/claude-agent-acp/blob/08e0e7f14ee1e221626a436b54b46ca32f2fe242/src/acp-agent.ts#L2682-L2690

The terminal result handler then forwards `message.result` whenever the active turn is a local-only command:

https://github.com/agentclientprotocol/claude-agent-acp/blob/08e0e7f14ee1e221626a436b54b46ca32f2fe242/src/acp-agent.ts#L3383-L3395

Although `local_command_output` marks top-level assistant text as delivered through `sendUpdate`, the `isLocalOnlyCommand` side of this condition bypasses the `deliveredAssistantText` guard:

```ts
session.activeTurn?.isLocalOnlyCommand ||
  (!deliveredAssistantText && (message.usage.output_tokens ?? 0) === 0)
```

Deduplicating arbitrary text in the ACP client would be unsafe because a model can legitimately repeat text, and the SDK event provenance has already been lost at that boundary. The fix should remain scoped to local-command output within the originating turn.

## Acceptance criteria

- When matching output arrives through both `local_command_output` and the terminal `result`, emit it once.
- When only `local_command_output` is present, emit it once.
- When only the terminal `result` is present, emit it once.
- Do not globally deduplicate ordinary assistant text.
- Do not suppress the same local command or output in a later turn.
- Add regression coverage for the dual-source and result-only cases.

## Related

- #432 introduced terminal-result forwarding for local-only slash commands.
- #858 added guarded result-text fallback for turns whose answer was not otherwise delivered.
- #349 tracks missing `/context` rendering, which is the opposite failure mode and should remain fixed.

ran411285752:
## Summary

- Deduplicate local-only slash command output at the SDK-to-ACP adapter boundary.
- Cover both SDK sequences: `local_command_output -> result` and `assistant -> result`.
- Keep evidence scoped to the originating turn or SDK cycle so ordinary assistant text and later turns are not globally deduplicated.

## Root cause

Claude Code can expose the same local command text in an intermediate SDK message and again in the terminal `result`. The adapter forwarded both as ACP `agent_message_chunk` notifications, after which clients could no longer distinguish their sources.

## Implementation

- Canonicalize only newline representation and trailing line feeds for local-command comparisons.
- Retain source-identified `local_command_output` evidence on its `Turn`.
- Retain consolidated assistant evidence only for the matching SDK result cycle.
- Clear cycle evidence on terminal results and authoritative non-lagged idle boundaries.
- Preserve result-only output, distinct output, repeated output in later turns, held-turn handoff, orphan lanes, and cancelled cycles.

## Validation on current main

Merged current `main` at `caf609b` and resolved the consumer/title-state overlap.

### Live end-to-end verification

Ran the built adapter as an ACP subprocess using the real local Claude login, created a real ACP session, and sent `/context` through `session/prompt`:

```text
process exit: 0
agent_message_chunk count: 1
Context Usage heading count: 1
end_turn count: 1
```

### Automated verification

- focused assembled-output/background-subagent selection: 57 passed;
- `npm run build`: passed;
- `npm run lint`: passed;
- `npx prettier --check src/acp-agent.ts src/tests/acp-agent.test.ts`: passed;
- `git diff --check`: passed;
- `npm run test:run`: 958 passed, 22 failed, 27 skipped. The failures are in untouched Windows path/permission display assertions and two 5-second timeout tests; all focused tests pass.

Fixes #948


ran411285752:
Merged current main at caf609b and resolved the consumer/title-state conflict. I also reran a real authenticated ACP subprocess test of /context: exit 0, one agent_message_chunk, one Context Usage heading, and one end_turn. Focused coverage is green (57 tests), with build/lint/changed-file Prettier passing. @benbrandt, could you please approve Actions and review the updated branch?

benbrandt:
closing in favor of #1078

ran411285752:
@benbrandt Thanks for taking a look. I think #949 and #1078 address two separate cases. #1078 gives local-command output its own message boundary so it does not run into a subsequent model reply. #949 addresses duplicate output *within the same local-only command turn*: when the SDK supplies the command body in an `assistant` (or `local_command_output`) message and again in the terminal `result`, ACP sends the body twice as `agent_message_chunk` updates.

On current main, `shouldForwardResult` is true whenever `session.activeTurn?.isLocalOnlyCommand` is true, even if that text was already delivered. I reproduced two identical ACP notifications for an `/context`-shaped `assistant -> result` sequence; this is not just a client rendering/concatenation issue. The #1078 diff does not change that result-forwarding path, so it would not address this duplication.

Would you be willing to reopen #949 to review it as a separate fix? I'm happy to adjust or narrow the implementation if you would prefer a different approach.

Issue #927, opened 2026-07-27
Expose monotonic provider token counters during active ACP turns


olympum:
## Problem

`@agentclientprotocol/claude-agent-acp` 0.63.0 exposes two useful usage channels, but neither supports a live token ceiling during `session/prompt`.

Standard `usage_update.used` reports current context occupancy. It is a gauge, so compaction can reduce it even though provider token spend cannot decrease.

Experimental `PromptResponse.usage` reports provider counters after the turn completes. Those counters are suitable for completed-turn accounting, although they arrive too late for a client to cancel an over-budget turn.

A captured authenticated 0.63.0 turn showed both channels:

```json
{"sessionUpdate":"usage_update","used":39486,"size":1000000}
{"stopReason":"end_turn","usage":{"inputTokens":4,"outputTokens":487,"cachedReadTokens":57378,"cachedWriteTokens":20760,"totalTokens":78629}}
```

The values diverged before compaction because they measure different properties.

## Request

Please expose provider-grounded token counters while a prompt is active. A namespaced `usage_update._meta` extension would work, although a standard ACP field would also solve the problem.

The signal needs these properties:

- Report provider counters rather than traffic-derived estimates.
- Include `inputTokens` and `outputTokens`, plus cache counters when Claude supplies them.
- Update after each completed provider result inside an agent loop.
- Declare whether counters are turn-scoped or session-scoped.
- Never decrease within the declared scope when compaction removes context.
- Define the reset boundary when a new prompt starts.
- Preserve the latest observed counters when a turn is cancelled.

The adapter already accumulates provider usage while a prompt runs, so this request is about exposing that live state rather than deriving a new estimate.

## Use case

Aegis enforces a per-invocation provider-token ceiling. Its client must send `session/cancel` when measured spend crosses the threshold.

Context occupancy cannot drive that decision because cache and compaction change occupancy independently of spend. Final `PromptResponse.usage` cannot drive it because the cancellation point has already passed.

## Compatibility

This request does not change the current completed-turn contract. `PromptResponse.usage` can remain per-turn metadata, consistent with #390.

It also complements #844: reporting usage on the cancelled response preserves final accounting, while this signal lets the client decide when to cancel.

Issue #914, opened 2026-07-25
Surface the advisor tool over ACP: result dropped, no config option, advisor tokens missing from usage


Seluj78:
**Note:** Claude handled the investigation and prose via back-and-forth. Idea, decisions, and everything else are mine. — @Seluj78

## Problem

The [advisor tool](https://platform.claude.com/docs/en/agents-and-tools/tool-use/advisor-tool) pairs a cheap executor model with a stronger advisor model that gives strategic guidance mid-turn. Claude Code supports it natively (`--advisor <model>`, `/advisor`, the `advisorModel` setting), and the adapter loads user settings, so an ACP client whose user has set `advisorModel` gets advisor calls today.

Over ACP, those calls are close to invisible:

1. The `advisor_tool_result` content block is dropped, so the tool call never gets a terminal status and the advice text is never delivered.
2. There is no advisor `config_option`, so a client cannot show or change the advisor model the way it does model / mode / effort / fast / agent.
3. Advisor tokens never reach `usage_update`, so a client's cost display understates an advisor-enabled session, which is precisely the number a user enabling this feature to save money is watching.

Observed with `@agentclientprotocol/claude-agent-acp@0.62.0`. Client is [Agent of Empires](https://github.com/agent-of-empires/agent-of-empires), whose structured view renders ACP sessions natively.

## 1. `advisor_tool_result` is dropped

`server_tool_use` shares the normal tool-call path, so the advisor consultation does reach the client as a `tool_call` named `advisor` with empty `input` (`dist/acp-agent.js:5469`, and per the API docs `server_tool_use.input` is always empty by design). The matching result is then discarded (`dist/acp-agent.js:5698-5701`):

```js
case "compaction_delta":
case "advisor_tool_result":
case "mid_conv_system":
case "fallback":
    break;
```

Two consequences for the client:

- No `tool_call_update` is emitted for that `tool_use_id`, so the card has no terminal status. It renders as a tool call that started and never finished.
- The advice is unavailable even when it is plaintext. Per the docs, `advisor_tool_result.content` is `advisor_result` (with `text`) for advisors like `claude-opus-4-8`, and `advisor_redacted_result` (encrypted) for Opus 5 / Fable 5 / Mythos 5.

[...]

Issue #844, opened 2026-07-07
Expose accumulated token usage on cancelled PromptResponse


yuva-dev:
### Summary

A cancelled turn currently resolves to `{ stopReason: "cancelled" }` with no `usage` field, whereas the `end_turn` and `refusal` settles include `usage: sessionUsage(session)`. Since `session.accumulatedUsage` already tracks the turn's token counts across every round-trip that completed before the interrupt, would you consider exposing it on the cancelled `PromptResponse` too?

### Why it would help

- A cancelled turn still consumes tokens — the input/output/cache from the round-trips that completed before the cancel are already in `session.accumulatedUsage`, but that data is dropped when the turn settles cancelled.
- Clients that meter or bill token usage (e.g. per-user budgets/quotas) can't account for a cancelled turn's spend. A user can cancel just before completion and those tokens become invisible to the client.
- There's no other channel for it client-side: per-turn token counts arrive only on the result; `usage_update.used` is context occupancy rather than consumption, and `cost` is a USD amount, not token counts.
- It would make the cancelled settle symmetric with the `end_turn`/`refusal` branches that already return `usage: sessionUsage(session)`.

### I realise the current behaviour is intentional

The cancelled settles deliberately resolve to exactly `{ stopReason: "cancelled" }`, and there are several assertions (`toEqual({ stopReason: "cancelled" })`) across the cancel / wedge (#680) / orphan (#825) edge cases locking that in — so I wanted to ask before proposing a change rather than push a PR that breaks those.

### Questions

1. Would you accept exposing the accumulated usage on cancelled `PromptResponse` (symmetric with `end_turn`)?
2. If so, any guidance on which settle points should carry it? The active-turn cancels (force-cancel/wedge, stream-end, idle-without-result, result-while-cancelled) each hold a `session.accumulatedUsage` that belongs to the cancelled turn — but the queued-turn `turn.resolve` in `cancel()` settles turns that never ran, so `sessionUsage(session)` wouldn't belong to those.

Happy to open a PR (with the corresponding test updates) if this is something you'd take.

Related: #390 (per-turn vs cumulative `PromptResponse.usage`).

Issue #390, opened 2026-03-07
PromptResponse.usage sends per-turn token counts, RFD says cumulative


hancengiz:
Hey, I'm building an ACP client ([Fabriqa](https://fabriqa.dev)) and noticed that `PromptResponse.usage` token counts don't match what the [session-usage RFD](https://agentclientprotocol.com/rfds/session-usage#session-usage-and-context-status) specifies.

The RFD defines the fields as cumulative across the session:
- `input_tokens`: "Total input tokens across all turns"
- `output_tokens`: "Total output tokens across all turns"

But claude-agent-acp resets `accumulatedUsage` to zero at the start of each `prompt()` call ([acp-agent.ts L488-493](https://github.com/zed-industries/claude-agent-acp/blob/main/src/acp-agent.ts#L488-L493)), so it sends per-turn values instead.

**Session 1** — two simple messages:

| Turn | inputTokens | outputTokens | cachedWriteTokens | cachedReadTokens | totalTokens |
|------|------------|-------------|------------------|-----------------|-------------|
| 1 | 3 | 22 | 42,653 | 0 | 42,678 |
| 2 | 3 | 49 | 30 | 42,653 | 42,735 |

**Session 2** — "how are you?" then just "a":

| Turn | inputTokens | outputTokens | cachedWriteTokens | cachedReadTokens | totalTokens |
|------|------------|-------------|------------------|-----------------|-------------|
| 1 | 3 | 20 | 5,740 | 36,912 | 42,675 |
| 2 | 3 | 17 | 25 | 42,652 | 42,697 |

If cumulative, turn 2 `inputTokens` should be 6, `outputTokens` should grow, etc. Instead each turn's `totalTokens` = sum of its own fields — clearly per-turn.

`usage_update.cost.amount` IS cumulative as the RFD specifies (Session 1: $0.267 → $0.290, Session 2: $0.055 → $0.077). So it's only `PromptResponse.usage` that differs from the spec.

For now in Fabriqa we treat `PromptResponse.usage` as per-turn and accumulate our own running total. Would be good to know — is the RFD going to change to match the current behavior, or should claude-agent-acp be updated to send cumulative values?

hancengiz:
<img width="3170" height="3558" alt="Image" src="https://github.com/user-attachments/assets/261d4293-4af3-4821-ab26-e7a88773cdc7" />
screenshot does not show the exact same values from the "session 2" example, I started another session with acp debug logs enabled but show the same finding.
[...]

PR #954, opened 2026-08-03
fix(accounting): report deduplicated SDK result usage

### LINKED ISSUE #953: Add deduplicated token accounting for failed and autonomous SDK results
## Problem

ACP Hosts currently cannot reconcile Claude provider token usage across all SDK result lifecycles without changing the existing per-turn meaning of `PromptResponse.usage`.

The missing accounting falls into two categories:

1. Failed SDK results contain non-zero `usage`, `modelUsage`, cost, and a stable UUID, but the prompt is rejected after the result is processed, so the Host receives no terminal usage value.
2. Subagent, background, and autonomous model work can be absent from the parent result's top-level `usage`. The SDK's `modelUsage` contains the aggregate provider total, but it is a session-cumulative snapshot rather than an additive per-result value.

Standard `usage_update.used` is context-window occupancy, not model consumption, and should not be repurposed.

## Scope

Add a negotiated Claude extension notification for provider accounting while preserving existing ACP semantics:

- Do not change `PromptResponse.usage`.
- Do not change standard `usage_update`.
- Emit at most one accounting event for each unique SDK result UUID.
- Emit before error rejection, cancellation exits, autonomous exits, and normal settlement.
- Include successful, failed, cancelled-result, background, subagent, and autonomous work.
- Derive additive usage from consecutive aggregate `modelUsage` snapshots.
- Keep raw result usage and cumulative model usage as non-additive audit fields.
- Require explicit Host capability negotiation.

## Proposed contract

Host capability:

```json
{
  "clientCapabilities": {
    "_meta": {
      "claudeCode": {
        "accountingUsage": { "version": 1 }
      }
    }
  }
}
```

Agent notification:

```text
_claude/accountingUsage
```

Consumer rules:

- Deduplicate by `resultId`.
- Add only `params.usage`.
- Do not add `PromptResponse.usage`, `resultUsage`, or raw cumulative `modelUsage`.
- Mark fallback and cumulative-snapshot reset semantics explicitly.

## Verified SDK behavior

Using Claude Agent SDK 0.3.220:

- Aggregate `modelUsage` is monotonic across ordinary turns, subagents, background tasks, autonomous results, model switches, and compaction.
- Same-model and cross-model subagent usage appears once in the aggregate snapshot delta.
- Error and post-cancel results can retain non-zero usage and stable UUIDs.
- A `/compact` result can have zero top-level `usage` while cumulative `modelUsage` still advances.
- Transcript APIs and JSONL may prune or omit compact-call model records, so transcript history is not a complete accounting ledger after compaction.

## Acceptance criteria

- Every unique SDK result produces zero or one negotiated accounting event.
- Failed result usage is emitted before the JSON-RPC error.
- Autonomous and background usage is emitted without entering the active user's `PromptResponse.usage`.
- Received results after cancellation retain provider accounting.
- Four token buckets are preserved: input, output, cache read, and cache creation.
- Consecutive cumulative snapshots are differenced rather than summed.
- Snapshot reset, invalid snapshot, and fallback recovery do not double count.
- Accounting event UUIDs reconcile one-to-one with SDK result UUIDs.
- Provider accounting totals reconcile with observable model-call totals where the SDK transcript exposes them.

## Related issues and PRs

- #927 covers monotonic provider counters during active turns; this issue covers terminal result accounting across failure and autonomous lifecycles.
- #914 documents auxiliary/advisor model usage missing from top-level usage.
- #844 / #846 cover cancelled-turn `PromptResponse.usage`, not provider result accounting.
- #870 covers background-subagent lifecycle while preserving per-turn PromptResponse semantics.
- #390 discusses retaining per-turn PromptResponse usage semantics.
- anthropics/claude-agent-sdk-typescript#112
- anthropics/claude-agent-sdk-typescript#293

ran411285752:
## Purpose

Preserve provider token accounting for failed, cancelled-result, autonomous, background, and subagent work without changing existing ACP turn-usage semantics.

Closes #953.

Related to #927 and #914.

## Protocol impact

- Adds a negotiated `_claude/accountingUsage` version 1 extension notification.
- Advertises support through `agentCapabilities._meta.claudeCode.accountingUsage`.
- Requires the Host to opt in with `clientCapabilities._meta.claudeCode.accountingUsage.version = 1`.
- Does not change standard `usage_update`.
- Does not change `PromptResponse.usage`.

## Accounting semantics

- `resultId` is the stable deduplication key.
- Only `params.usage` is additive.
- `modelUsage` is a cumulative SDK audit snapshot and must not be summed directly.
- `resultUsage` is diagnostic top-level result usage and must not be added to `params.usage`.
- `PromptResponse.usage` and accounting events must not be summed together.
- Consecutive aggregate model snapshots are differenced after aggregating all model keys.
- Invalid snapshots use an explicit fallback source.
- Fallback usage is reconciled when cumulative snapshots resume, avoiding double counting.
- Snapshot resets start a new accounting epoch.

## Lifecycle coverage

Accounting is attempted before:

- failed result rejection;
- cancellation exits;
- autonomous result exits;
- normal result settlement.

Optional extension delivery failures are logged without changing the standard result lifecycle.

## Evidence

Built-Adapter ACP subprocess scenarios covered:

- ordinary multi-turn usage;
- same-model and cross-model subagents;
- background and autonomous results;
- error results;
- cancellation after non-zero assistant usage;
- resume;
- compaction;
- model switching.

A-H and J:

```text
ACP accounting:             304,617
Unique public model ledger: 304,617
Difference:                       0
```

Compaction scenario I:

```text
Final cumulative modelUsage: 194,914
ACP accounting sum:          194,914
Difference:                        0
```

The `/compact` result had zero `resultUsage` and a 32,903-token accounting delta. Public transcript APIs and JSONL omit that compact-call usage, so they cannot be used as a complete post-compaction accounting ledger.

## Verification

- `npm run build`
- `npm run lint`
- Prettier check for all five changed files
- 78 focused accounting/lifecycle tests
- Full suite: 711 passed, 20 skipped
- Six pre-existing Windows path-separator assertions remain failing, matching clean `origin/main`
- Mandatory review: no Critical, High, or Medium findings

## Related context

- #927
- #914
- #844 / #846
- #870
- #390
- anthropics/claude-agent-sdk-typescript#112
- anthropics/claude-agent-sdk-typescript#293


ran411285752:
@benbrandt Could you confirm the preferred scope before I resolve the current main conflicts and re-run the evidence suite against Claude Agent SDK 0.3.238?

Would maintainers prefer:

1. **One result-accounting extension (current design):** emit one deduplicated event for every SDK result, including user-turn failures and autonomous/background results, while leaving PromptResponse.usage and standard usage_update unchanged; or
2. **Two staged changes:** first land deduplicated accounting for user-turn/success/error results, then add autonomous/background result reporting in a follow-up PR?

The key trade-off is that splitting reduces the first review surface, but the first PR would not yet close #953's missing background/autonomous accounting. I will hold off on conflict resolution and SDK 0.3.238 revalidation until the intended boundary is confirmed.

PR #964, opened 2026-08-06
feat: expose provider-neutral ACP goal extension

nikita-ashihmin:
## Summary

- advertise a provider-neutral goal extension in `InitializeResponse._meta.goal`
- map Claude SDK `active_goal` messages to `session_info_update._meta.goal`
- expose the shared `_session/goal` control method with the truthfully supported `clear` action
- document the extension as a candidate shape for a future first-class ACP goal API

## Architecture

A goal is persistent **session state**, not the lifecycle of one `session/prompt` request. `goal.status === "active"` says that the objective may drive more work; it does not say that a prompt is executing now.

This keeps two independent axes:

1. **Goal state** is published asynchronously as session metadata and may survive multiple prompt turns.
2. **Prompt activity** follows the backend's actual running/quiescent boundaries. While a turn runs, clients use advertised steering or prompt queueing. Once the backend is quiescent, an ordinary `session/prompt` is valid even if the goal remains active.

The metadata is deliberately provider-neutral:

```json
{
  "_meta": {
    "goal": {
      "version": 1,
      "controlMethod": "_session/goal",
      "actions": ["clear"]
    }
  }
}
```

State is emitted as `session_info_update._meta.goal`; `null` clears it. The common snapshot vocabulary supports `objective`, common statuses, timestamps in Unix milliseconds, budgets/usage, iteration count, and a continuation reason. Each implementation emits only fields it actually knows. This shape is intentionally close to a future typed ACP capability/update so promotion out of `_meta` would not require a provider-specific migration.

Control actions are capability-negotiated. The shared vocabulary reserves `set`, `pause`, `resume`, and `clear`, but an agent advertises only the subset it can implement with correct semantics.

## Claude mapping

The Claude runtime already emits internal `active_goal` messages, although that message is not currently included in the public `SDKMessage` union. The adapter handles it before the exhaustive switch:

- `condition` → `objective`
- `iterations` → `iterations`
- `set_at` → `createdAt`
- `last_reason` → `lastReason`
- `value: null` → `goal: null`

Provider bookkeeping such as `tokens_at_start` is intentionally not exposed. Goal updates do not settle, extend, or otherwise mutate ACP prompt lifecycle. Claude's existing Stop hook remains authoritative for whether the backend itself continues working.

Only `clear` is advertised for out-of-band control. It submits `/goal clear` through the existing session prompt queue, preserving the adapter's established ordering and turn ownership. Setting a goal remains available through `/goal <condition>` but is not advertised as a control action because it starts the long-running Stop-hook workflow.

The companion Codex implementation is agentclientprotocol/codex-acp#371.

## Verification

- `npm run test:run` — 692 passed, 20 skipped
- `npm run build`
- `npm run check`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -77,6 +77,7 @@ import {
   Query,
   query,
   SDKAssistantMessageError,
+  SDKActiveGoalMessage,
   SDKMessage,
   SDKMessageOrigin,
   SDKPartialAssistantMessage,
@@ -84,6 +85,14 @@ import {
   SlashCommand,
   ThinkingConfig,
 } from "@anthropic-ai/claude-agent-sdk";
+import {
+  GOAL_CONTROL_METHOD,
+  GOAL_EXTENSION_VERSION,
+  GoalControlRequest,
+  GoalControlResponse,
+  parseGoalControlRequest,
+  toGoalSnapshot,
+} from "./goal-extension.js";
 import { ContentBlockParam } from "@anthropic-ai/sdk/resources";
 import { BetaContentBlock, BetaRawContentBlockDelta } from "@anthropic-ai/sdk/resources/beta.mjs";
 import { execFile } from "node:child_process";
@@ -1596,6 +1605,11 @@ export class ClaudeAcpAgent {
         steering: {
           supported: true,
         },
+        goal: {
+          version: GOAL_EXTENSION_VERSION,
+          controlMethod: GOAL_CONTROL_METHOD,
+          actions: ["clear"],
+        },
       },
     };
   }
@@ -1892,6 +1906,14 @@ export class ClaudeAcpAgent {
     session.input.push(userMessage);
     this.ensureConsumer(session, params.sessionId);
     return response;
+  }
+
+  async controlGoal(params: GoalControlRequest): Promise<GoalControlResponse> {
+    await this.prompt({
+      sessionId: params.sessionId,
+      prompt: [{ type: "text", text: "/goal clear" }],
+    });
+    return {};
   }
 
   /** Steer the session per the ACP steering wire protocol: inject a follow-up
@@ -2632,6 +2654,21 @@ export class ClaudeAcpAgent {
               }
               break;
           }
+          continue;
+        }
+
+        // `active_goal` is emitted by the Claude runtime but is not currently
+        // included in the public SDKMessage union. Handle it before the
+        // exhaustive switch and publish only the provider-neutral ACP shape.
+        if ((message as { type: string }).type === "active_goal") {
+          const activeGoal = message as unknown as SDKActiveGoalMessage;
+          await this.client.sessionUpdate({
+            sessionId: params.sessionId,
+            update: {
+              sessionUpdate: "session_info_update",
+              _meta: { goal: toGoalSnapshot(activeGoal) },
+            },
+          });
           continue;
         }
 
@@ -7878,6 +7915,11 @@ export function runAcp() {
     .onNotification(methods.agent.session.cancel, (ctx) => agent.cancel(ctx.params))
     .onRequest<SteerRequest, SteerResponse>(STEER_METHOD, { parse: parseSteerRequest }, (ctx) =>
       agent.steer(ctx.params),
+    )
+    .onRequest<GoalControlRequest, GoalControlResponse>(
+      GOAL_CONTROL_METHOD,
+      { parse: parseGoalControlRequest },
+      (ctx) => agent.controlGoal(ctx.params),
     )
     .connect(stream);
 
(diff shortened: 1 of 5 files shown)


src/acp-agent.ts
```diff
@@ -1894,6 +1908,14 @@ export class ClaudeAcpAgent {
     return response;
   }
 
+  async controlGoal(params: GoalControlRequest): Promise<GoalControlResponse> {
```
EugeneTheDev: Let's call it just `goal`, to make it consistent with existing methods such as `steeer`

src/goal-extension.ts
```diff
@@ -0,0 +1,63 @@
...
+  lastReason?: string | null;
+  createdAt?: number;
+  updatedAt?: number;
+  tokenBudget?: number | null;
+  tokensUsed?: number;
+  timeUsedSeconds?: number;
+  controlMethod: typeof GOAL_CONTROL_METHOD;
+};
+
+export type GoalControlRequest = {
```
EugeneTheDev: Similar here, let's just call it `GoalRequest`

src/goal-extension.ts
```diff
@@ -0,0 +1,63 @@
+import { RequestError } from "@agentclientprotocol/sdk";
+import type { SDKActiveGoalMessage } from "@anthropic-ai/claude-agent-sdk";
+
+export const GOAL_EXTENSION_VERSION = 1 as const;
+export const GOAL_CONTROL_METHOD = "_session/goal";
+
+export type GoalControlAction = "set" | "pause" | "resume" | "clear";
+
+export type GoalCapability = {
```
EugeneTheDev, line 10: This capability record appears to be unused. I guess the intended use was to add it to `_meta` in the initialize request, to advertise goal capability. Can you please do this?

src/goal-extension.ts
```diff
@@ -0,0 +1,63 @@
+import { RequestError } from "@agentclientprotocol/sdk";
+import type { SDKActiveGoalMessage } from "@anthropic-ai/claude-agent-sdk";
+
+export const GOAL_EXTENSION_VERSION = 1 as const;
+export const GOAL_CONTROL_METHOD = "_session/goal";
+
+export type GoalControlAction = "set" | "pause" | "resume" | "clear";
```
EugeneTheDev: Here you list a number of values, but below, in the `parseGoalControlRequest` you explicitly throw if the value is not `clear`. That means only `clear` is actually supported, let's not advertise unsupported values. 

Also, let's rename it to `GoalAction`

src/goal-extension.ts
```diff
@@ -0,0 +1,63 @@
...
+};
+
+export type GoalControlRequest = {
+  sessionId: string;
+  action: "clear";
+};
+
+export type GoalControlResponse = Record<string, never>;
+
+export function parseGoalControlRequest(params: unknown): GoalControlRequest {
```
EugeneTheDev: Let's rename it to `parseGoalRequest` too

src/goal-extension.ts
```diff
@@ -0,0 +1,63 @@
...
+  if (!params || typeof params !== "object") {
+    throw RequestError.invalidParams(undefined, "goal params must be an object");
+  }
+  const { sessionId, action } = params as Record<string, unknown>;
+  if (typeof sessionId !== "string" || sessionId.length === 0) {
+    throw RequestError.invalidParams(undefined, "goal params require a non-empty sessionId");
+  }
+  if (action !== "clear") {
```
EugeneTheDev: Can you reuse `GoalControlAction` here, to make sure supported action values stay consistent with what's advertised

EugeneTheDev (changes requested):

src/acp-agent.ts
nikita-ashihmin: Fixed in bcb25b1: renamed the handler to goal.

src/goal-extension.ts
nikita-ashihmin: Fixed in bcb25b1: renamed the request type to GoalRequest.

src/goal-extension.ts
nikita-ashihmin: Fixed in bcb25b1: initialize advertises _meta.goal, and that wire object is now explicitly checked with satisfies GoalCapability.

src/goal-extension.ts
nikita-ashihmin: Fixed in bcb25b1: GoalAction is derived from the sole supported clear action.

src/goal-extension.ts
nikita-ashihmin: Fixed in bcb25b1: renamed to parseGoalRequest.

src/goal-extension.ts
nikita-ashihmin: Fixed in bcb25b1: request parsing and advertised capabilities now share GOAL_ACTIONS/GoalAction.

EugeneTheDev (approved):

PR #965, opened 2026-08-06
ci: simplify the release flow and make it agent-friendly

Rizzen:


CLAUDE.md
```diff
@@ -0,0 +1,2 @@
+See [AGENTS.md](AGENTS.md) — it is the single source of agent instructions for this
```
EugeneTheDev: You can just use a symlink to `AGENTS.md` here

scripts/release-preflight.sh
```diff
@@ -0,0 +1,103 @@
+#!/usr/bin/env bash
```
EugeneTheDev, line 1: Where is this file used? Is it used as a PR gate?

EugeneTheDev (approved):

scripts/release-preflight.sh
Rizzen: Yes, it's the script to verify if release flow is not broken. Mostly for agent to verify if everything is right

PR #969, opened 2026-08-07
fix(steering): reject a missing post-echo result

ran411285752:
## Summary

Follow-up to #958. After merging current `main` (`caf609b`), #967 already covers the pre-echo diagnostic-result failure, so this PR is now limited to the remaining post-echo result-boundary bug:

- clear the interrupted cycle's saved `steeredSettle` when the matching steering echo replays;
- if the SDK then goes idle without a result for the steered cycle, reject with the existing `no_result` error instead of reusing the interrupted cycle's successful result.

## User impact

A steered instruction that starts but never produces its own final result is no longer reported as successfully completed using the result from before the instruction ran.

## Tests

Fresh RED on current `main` behavior:

- `does not reuse the interrupted result after the steered echo` incorrectly resolved `end_turn` with 15-token pre-echo usage.

GREEN and adjacent verification:

- focused Steering/ExitPlanMode tests: 6 passed;
- `npm run build`: passed;
- `npm run lint`: passed;
- `npx prettier --check src/acp-agent.ts src/tests/acp-agent.test.ts`: passed;
- `git diff --check`: passed.

Windows repo-wide baseline:

- `npm run test:run`: 949 passed, 23 failed, 27 skipped. Failures are in untouched path/permission formatting tests and three 5-second timeouts; the focused Steering tests pass.
- repo-wide Prettier reports CRLF drift in 75 untouched files; both changed files pass direct Prettier checking.


ran411285752:
@Rizzen, could you please review this focused follow-up to #958 when you have a chance?

It keeps #958's steering state machine and only fixes the pre-echo diagnostic-result rejection and the post-echo stale-result reuse described above. The branch is mergeable and the focused Steering/background-subagent tests pass locally.

ran411285752:
Merged current main at caf609b and resolved the overlap with #967: this PR now contains only the post-echo stale-result fix. Fresh RED reproduced the false end_turn success; the focused Steering/ExitPlanMode suite is green (6 tests), and build/lint pass. @Rizzen, could you please approve Actions and review the narrowed diff?

benbrandt:
It seems this was fixed in #1166

Issue #976, opened 2026-08-09
Follow-up prompts kill the background subagents a held turn is waiting for


Sergio-3:
## Summary

Since #870, a turn whose terminal result arrives while background subagents it spawned are still live is held open — its `session/prompt` stays pending. This makes an ordinary follow-up message lethal to those subagents on clients that cancel before each prompt: Zed's `run_turn` sends `session/cancel` unconditionally whenever a turn appears to be running (`crates/acp_thread/src/acp_thread.rs`, `run_turn` → `cancel_inner`), and a held turn *always* appears to be running. `cancel()` then calls `query.interrupt()`, tearing down the very subagents the hold exists to serve — the comment above the held-turn settle in `cancel()` states it itself: *"…its background subagents, which the interrupt below tears down."*

The subagents die mid-work and their completion notifications are lost — the model never learns the task finished.

Observed on 0.64.2; verified still present on current `main` (`1afa940`): `cancel()` interrupts unconditionally.

## Observed (long session, ~24 subagent jobs, Zed 1.14.2)

- Background subagents (Agent tool, `run_in_background: true`) reliably died the moment a new user message arrived — no ESC, no explicit cancel.
- Partially written output files on disk prove they were mid-work.
- A subagent that *completed successfully* had its completion notification dropped: the output file existed, the wake-up never happened. (Its `task_notification` would have triggered a follow-up turn inside the held turn; the interrupt killed that follow-up before it ran.)
- Background *shells* were unaffected throughout — consistent with the code: they never defer a turn, so no cancel ever reaches them.

## Root cause chain

1. Turn spawns a background subagent → result arrives → turn is parked (`Turn.deferredSettle`), `session/prompt` stays pending.
2. Client sees an unresolved prompt → status "generating" → next user message triggers the client's routine pre-prompt cancel.
3. `cancel()` settles the held turn, then `await session.query.interrupt()` — killing the subagents and any pending task-notification follow-up.
4. The `session.cancelled` latch additionally makes the consumer drop every message until the next turn activates — so even output from a subagent that survived the interrupt window is discarded.

## Secondary symptom (same root)

While a turn is held, the client cannot distinguish "orchestrator idle, subagents working" from "model generating". Zed queues the user's next message instead of sending it, which is exactly backwards during orchestration: the main loop is idle and could accept input, yet the UI signals it is busy.

## The hold's rationale does not apply to Zed (verified empirically)

[...]

PR #977, opened 2026-08-09
fix: keep background subagents alive across a follow-up cancel

Sergio-3:
Implements the minimal fix (a) proposed in #976.

Since #870, a turn whose terminal result arrives while the background subagents it spawned are still live is held open, so its `session/prompt` stays pending. Clients that cancel before each prompt therefore reach `cancel()` on every ordinary follow-up message — Zed's `run_turn` cancels unconditionally whenever a turn appears to be running, and a held turn always does — and `query.interrupt()` tears down the very subagents the hold exists to serve. Their work is lost, and the `task_notification` that would have reported completion never fires.

This skips the interrupt in the one state where there is provably nothing to interrupt: the turn settled just above was only held open, no other turn is active, orphaned or queued, and the session's last consumed state is `idle`. Every other shape still interrupts exactly as before.

Clearing the `cancelled` latch is part of the skip rather than an aside: while it is set the consumer drops every incoming message and only `activateTurn` clears it, so leaving it set would swallow the surviving subagents' output anyway — and permanently if no further prompt follows.

The resulting behaviour matches how background *shells* already work: they never defer a turn, so no cancel has ever reached them.

Tests:

- `keeps a hold's subagents alive across a cancel on a quiet session`
- `still interrupts a cancel during a hold whose session is running`

Verified against Zed across a long orchestration session: background subagents now survive follow-up messages, and their completion notifications arrive. The structural option (b) in #976 — releasing the hold once the session goes idle — is intentionally not part of this PR.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -4803,10 +4803,12 @@ export class ClaudeAcpAgent {
     // trailer debt is absorbed by the idle handler when its idle does come.
     // The turn's own usage snapshot is reported per the cancelled-usage
     // contract (issue #844).
+    let settledHeldTurn = false;
     {
       const active = session.activeTurn;
       if (isHeldOpen(active)) {
         this.finishFileChangeAudit(session, active, "cancelled");
+        settledHeldTurn = true;
         active.settled = true;
         // Mirror settleActive's invariants (it is consumer-scoped and
         // unreachable from here): disarm the backstop — none should be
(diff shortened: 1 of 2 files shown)


kyle-see:
Another data point in favour of this fix, from a different platform — no action needed beyond a review whenever someone has a moment.

**Environment:** Windows 11, Zed 1.16.1, an orchestrator session with several background subagent lanes running.

**Symptom:** matches #976 exactly. Sending an ordinary follow-up message to the orchestrator tears down every in-flight subagent; they report "stopped by the user". No ESC, no stop button, no explicit cancel — the intent was only to talk to the orchestrator while it kept working.

Queueing the message isn't a workaround either: while the turn is held open for the background lanes, a queued message never releases.

Happy to test a build of this branch against a real multi-lane run on Windows if that would be useful. Thanks for putting the fix together.

benbrandt:
It looks like this fix no longer does what is expected on the latest SDK

PR #979, opened 2026-08-09
feat: expose typed session failures for AIR

nikita-ashihmin:
## Summary

- Negotiate the JetBrains AIR `sessionFailure` extension when the client advertises an integer version `>= 1` and the capability name.
- Return turn-terminal failures only on `PromptResponse._meta`; publish idle and asynchronous session health failures through `session/update`.
- Preserve legacy ACP behavior for clients that do not negotiate the extension.
- Emit bounded fixed safe messages without SDK/provider details, paths, credentials, or stack traces.

## Recovery contract

Failures use `_meta.jetbrains.air.sessionFailure` with a stable id, monotonic revision, phase, category, source, safe message, retryability, and optional real turn attribution. Session-scoped ids include a consumer epoch, so replayed tombstones from a dead consumer cannot suppress failures after restart. Stale consumers cannot publish into a replacement session.

AIR owns localized action presentation. Claude supplies typed category and retryability facts. Billing maps to quota exhaustion, max-output limits map to context exhaustion, and the irreversibly closed Claude Query reports non-retryable `transport_lost` so AIR offers New Session.

A held-open turn keeps its already recorded successful `PromptResponse`, while worker shutdown or transport death also emits a separate session-scoped health failure without `turnId`; the next prompt is proven to fail because the Query is closed. Worker-shutdown remains armed through buffered trailing frames and is classified only when EOF proves the live tail.

## Validation

- 750 passed, 25 skipped.
- Build passed.
- ESLint passed.
- Changed files pass Prettier; the repository-wide check is blocked only by an unrelated untracked local proposal file.
- Tests cover negotiation, taxonomy, sanitization, terminal/asynchronous carrier separation, stale consumers, worker replay/live-tail behavior, held turns, process/transport failure, restart-safe ids, clear revisions, and capability-off behavior.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -916,9 +916,124 @@ export type ToolUpdateMeta = {
 };
 
 const SUBAGENT_TRANSCRIPT_CAPABILITY = "subagent-transcript";
+const AIR_EXTENSION_VERSION = 1;
 
 function supportsSubagentTranscript(capabilities?: ClientCapabilities | null): boolean {
   return capabilities?._meta?.[SUBAGENT_TRANSCRIPT_CAPABILITY] === true;
+}
+
+type AirSessionFailureCategory =
+  | "auth_required"
+  | "bad_request"
+  | "budget_exhausted"
+  | "context_exhausted"
+  | "internal_error"
+  | "overloaded"
+  | "provider_error"
+  | "quota_exhausted"
+  | "rate_limited"
+  | "transport_lost"
+  | "worker_shutdown";
+
+const AIR_FAILURE_PRESENTATION: Record<
+  AirSessionFailureCategory,
+  { message: string; retryable: boolean; actions: string[] }
+> = {
+  auth_required: {
+    message: "Sign in to continue using Claude.",
+    retryable: false,
+    actions: ["login"],
+  },
+  bad_request: {
+    message: "Claude could not process this request.",
+    retryable: false,
+    actions: ["new_turn"],
+  },
+  budget_exhausted: {
+    message: "This Claude session reached its configured budget.",
+    retryable: false,
+    actions: ["new_session"],
+  },
+  context_exhausted: {
+    message: "This Claude turn reached its configured limit.",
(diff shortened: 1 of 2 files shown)


src/acp-agent.ts
```diff
@@ -916,11 +916,126 @@ export type ToolUpdateMeta = {
...
+    Number.isInteger(version) &&
+    version >= AIR_EXTENSION_VERSION &&
+    Array.isArray(advertised) &&
+    advertised.some((capability) => capability === "sessionFailure")
+  );
+}
+
+function providerFailureCategory(errorKind?: SDKAssistantMessageError): AirSessionFailureCategory {
+  switch (errorKind) {
+    case "authentication_failed":
```
Ololoshechkin, line 1025: nit: extract to constant varialbes

src/acp-agent.ts
```diff
@@ -1623,6 +1738,12 @@ export class ClaudeAcpAgent {
       // steering extension contract: advertises the `_session/steering` request
       // so clients know they may inject a follow-up into a running turn.
       _meta: {
+        jetbrains: {
+          air: {
```
Ololoshechkin: Do we really need to expose our product name in these fields?

Ololoshechkin (approved):

src/acp-agent.ts
nikita-ashihmin: Fixed in 830096f. The vendor key, extension namespace, version/capability keys, and sessionFailure key are now named constants used by both capability negotiation and metadata emission.

src/acp-agent.ts
nikita-ashihmin: The `air` segment is intentional protocol namespacing rather than user-facing branding: `jetbrains` owns this non-standard extension, while `air` identifies the client contract that defines the rendering semantics. Removing it would make the metadata apply to every JetBrains ACP client and increase collision/accidental-decoding risk. I documented that rationale next to the centralized wire constants in 830096f.

PR #984, opened 2026-08-11
feat: add ai generated session title

EugeneTheDev:
Claude Code's automatic title generation never runs under the Agent SDK, so `SDKSessionInfo.summary` degrades to the raw first prompt — meaning we were publishing up to 256 characters of the user's opening message as the session title.

This PR asks the SDK for a real title instead. On the first completed turn we send the `generate_session_title` control request with the trailing 1000 characters of the session's own user + assistant text, fired in the background off the `idle` handler so turn-end never waits on the ~2 s small-model call, and publish the result through the existing `session_info_update`. 

A title is generated once per session, and any title already in the session file — a user `/rename` or one we generated earlier, which the SDK folds into the same field — is adopted rather than overwritten.`persist: true` writes it to the transcript, so `session/list` and `session/load` show the same name.

The SDK method is undeclared in `sdk.d.ts`, so it's called behind a `typeof` check and a `try/catch` that degrades to the previous behaviour.

### DIFF
--- a/src/session-titles.ts
+++ b/src/session-titles.ts
@@ -1,0 +1,256 @@
+/**
+ * AI-generated session titles.
+ *
+ * Claude Code's own auto-title generation never runs under the Agent SDK — the
+ * latch that arms it is pre-set in the headless path — so `SDKSessionInfo.summary`
+ * degrades to the raw first prompt. This module asks the CLI for a real title
+ * instead, via the `generate_session_title` control request, and publishes it
+ * over ACP as a `session_info_update`.
+ *
+ * {@link SessionTitles} holds the per-session state and is owned by `Session`.
+ * `acp-agent.ts` drives it from four places: {@link SessionTitles.onPrompt} and
+ * {@link SessionTitles.onAssistantText} to collect text,
+ * {@link SessionTitles.onTurnEnd} at `session_state_changed: idle`, and
+ * {@link SessionTitles.reset} on `conversation_reset`.
+ */
+
+import type { ContentBlock, PromptRequest } from "@agentclientprotocol/sdk";
+import { getSessionInfo, type Query, type SDKSessionInfo } from "@anthropic-ai/claude-agent-sdk";
+import type { ClaudeAcpAgent, Session } from "./acp-agent.js";
+
+const MAX_TITLE_LENGTH = 256;
+
+/** How much session text the title is generated from. The CLI caps its own
+ *  title input at the same 1000 trailing characters. */
+const MAX_TITLE_CONTEXT_LENGTH = 1000;
+
+/** Below this the generator returns null without calling the model, so there is
+ *  nothing to gain by asking yet. */
+const MIN_TITLE_CONTEXT_LENGTH = 10;
+
+/** `generateSessionTitle` is present on the SDK's runtime `Query` class but is
+ *  not declared in `sdk.d.ts` (0.3.220), so it is reached through this optional
+ *  shape rather than the published interface. */
+type TitleCapableQuery = Query & {
+  generateSessionTitle?: (
+    description: string,
+    options?: { persist?: boolean },
+  ) => Promise<string | null>;
+};
+
+/** A title already stored for the session, published when generation is
+ *  unavailable or yields nothing. */
+type TitleFallback = { title: string; lastModified: number };
+
+export function sanitizeTitle(text: string): string {
+  // Replace newlines and collapse whitespace
+  const sanitized = text
+    .replace(/[\r\n]+/g, " ")
+    .replace(/\s+/g, " ")
+    .trim();
+  if (sanitized.length <= MAX_TITLE_LENGTH) {
+    return sanitized;
+  }
+  return sanitized.slice(0, MAX_TITLE_LENGTH - 1) + "…";
+}
+
+/** Rolling text a title is generated from: `previous` plus `text`, keeping only
+ *  the trailing {@link MAX_TITLE_CONTEXT_LENGTH} characters. */
+export function appendTitleContext(previous: string | undefined, text: string): string {
+  const next = (previous ?? "") + text;
+  return next.length > MAX_TITLE_CONTEXT_LENGTH ? next.slice(-MAX_TITLE_CONTEXT_LENGTH) : next;
+}
+
+/** Whether a prompt is worth titling a session after. The CLI's own auto-title
+ *  path skips slash commands, `!bash` lines and tagged system text; match it, or
+ *  sessions end up named "/compact". */
+function isTitleWorthyPrompt(text: string): boolean {
+  const trimmed = text.trim();
+  return trimmed.length > 0 && !/^[/!<]/.test(trimmed);
+}
+
+/** Whether this SDK still offers on-demand title generation. */
+function supportsTitleGeneration(query: Query): boolean {
+  return typeof (query as TitleCapableQuery).generateSessionTitle === "function";
+}
+
+/** One session's title: the text a generated title is derived from, whether a
+ *  title has been settled on, and the last one published. Created with the
+ *  session and reachable as `session.titles`. */
+export class SessionTitles {
+  /** Last title pushed to the client via `session_info_update`, so an unchanged
+   *  title is never re-notified. Undefined until the first push. */
+  private lastTitle?: string;
+
+  /** Rolling tail of this session's own user + assistant text, capped at
+   *  {@link MAX_TITLE_CONTEXT_LENGTH}. Dropped once a title exists. */
+  private context?: string;
+
+  /** Set once this session has a title or has asked for one. A title is
+   *  generated at most once per session: the SDK reports a generated title in
+   *  the same field as a user `/rename`, so re-titling could silently overwrite
+   *  one. Released by {@link reset}, and when generation yields nothing. */
+  private settled = false;
+
+  constructor(
+    private readonly agent: ClaudeAcpAgent,
+    private readonly sessionId: string,
+  ) {}
+
+  /** Collect a prompt's own text for the title, skipping openers not worth
+   *  titling after. No-op once the title is settled. */
+  onPrompt(prompt: PromptRequest["prompt"]): void {
+    if (this.settled) {
+      return;
+    }
+    const promptText = prompt
+      .flatMap((chunk) => (chunk.type === "text" ? [chunk.text] : []))
+      .join("\n");
+
+    if (isTitleWorthyPrompt(promptText)) {
+      this.context = appendTitleContext(this.context, `\n${promptText}\n`);
+    }
+  }
+
+  /** Collect the assistant's own answer for the title. Chunks are appended
+   *  verbatim so streamed text reassembles. No-op once the title is settled. */
+  onAssistantText(content: ContentBlock): void {
+    if (this.settled || content.type !== "text") {
+      return;
+    }
+    this.context = appendTitleContext(this.context, content.text);
+  }
+
+  /** Drop the title state so the next turn re-evaluates. Used on
+   *  `conversation_reset`, which mounts a fresh transcript. */
+  reset(): void {
+    this.settled = false;
+    this.context = undefined;
+    this.lastTitle = undefined;
+  }
+
+  /** Turn-end title handling. `idle` is the SDK's turn-over signal, so it is
+   *  when a title may have landed or become generatable.
+   *
+   *  `info.customTitle` is non-empty only once the session file holds a real
+   *  title — a user `/rename` or one generated earlier; the SDK folds both into
+   *  that one field. Adopt it and latch, so neither is ever titled over.
+   *
+   *  With no title yet, ask the SDK to generate one in the background: it is a
+   *  ~2s small-model call and turn-end must not wait on it. Only when that is
+   *  not possible do we fall back to `summary`, which for an SDK-driven session
+   *  is just the raw first prompt. */
+  async onTurnEnd(session: Session): Promise<void> {
+    const info = await this.readSessionInfo(session);
+
+    if (info?.customTitle) {
+      this.settled = true;
+      this.context = undefined;
+      await this.publish(info.customTitle, info.lastModified);
+      return;
+    }
+
+    const fallback = info?.summary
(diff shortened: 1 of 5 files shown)


OSadovy:
`isTitleWorthyPrompt` drops a slash command's arguments along with the command.

`/^[/!<]/` rejects the whole prompt, so a session opened with
`/investigate-alert kafka lag on prod-3` contributes nothing to the context and
the title ends up derived from the assistant's reply alone. For a skill
invocation the arguments are usually the only statement of what the session is
for.

The `/compact` case the comment cites is really about *argument-less* commands —
skipping when nothing follows the command token would keep it out while leaving
`/skill <args>` in.

This is somewhere the adapter can do better than the CLI rather than match it. The CLI has no access to these arguments: it persists an
invocation as `<command-name>`/`<command-args>` and its title selector skips
messages starting with those markers, so both the command and its arguments are
invisible to it. `onPrompt` sees the raw text before that wrapping.

src/session-titles.ts
```diff
@@ -0,0 +1,256 @@
...
+ *  the trailing {@link MAX_TITLE_CONTEXT_LENGTH} characters. */
+export function appendTitleContext(previous: string | undefined, text: string): string {
+  const next = (previous ?? "") + text;
+  return next.length > MAX_TITLE_CONTEXT_LENGTH ? next.slice(-MAX_TITLE_CONTEXT_LENGTH) : next;
+}
+
+/** Whether a prompt is worth titling a session after. The CLI's own auto-title
+ *  path skips slash commands, `!bash` lines and tagged system text; match it, or
+ *  sessions end up named "/compact". */
+function isTitleWorthyPrompt(text: string): boolean {
```
Ololoshechkin, line 67: @EugeneTheDev let's drop this one. I don't think it makes that much sense + in Codex we don't do this anyway. Let's do it same way at least

Ololoshechkin (approved):

tcztzy:
Hi, I investigated title-generation requests continuing after setting `CLAUDE_CODE_DISABLE_TERMINAL_TITLE=1`, and wanted to share the results here because they seem relevant to the rationale for this PR.

**Native automatic title generation ran in SDK-only sessions with both the historical SDK versions around this PR and the SDK currently pinned by main.** This happened in mock-backed tests and with real model responses. There may be a condition specific to the original setup that I have not reproduced, so I would appreciate any context on that.

**Versions and real API results**

I checked the dependency pins at these points:

- PR opened on August 11: [SDK `0.3.220`](https://github.com/agentclientprotocol/claude-agent-acp/blob/94fd152bb2a839bd989e8744252c07d914fb6c64/package.json), bundled Claude Code `2.1.220`.
- PR merged on August 19: [SDK `0.3.232`](https://github.com/agentclientprotocol/claude-agent-acp/blob/d334766ef95dd89201979d42252e3d2a5a259cb9/package.json), bundled Claude Code `2.1.232`.
- Main checked on September 8: [SDK `0.3.257`](https://github.com/agentclientprotocol/claude-agent-acp/blob/d870ebd2fd14bda95f81ec4b86d76ef4ae258005/package.json), bundled Claude Code `2.1.257`.

For the live tests, I called the official SDK's `query()` directly with asynchronous streaming input, a fresh session/config directory, `settingSources: []`, and `tools: []`. The ACP adapter was not loaded, and the harness never called `generateSessionTitle()` or sent `generate_session_title`.

The real provider was OpenCode Go (`https://opencode.ai/zen/go`), serving DeepSeek V4 Flash through its Anthropic-compatible endpoint. A local recorder forwarded the requests and actual response streams. It did not generate responses. The configured model was `deepseek-v4-flash[1m]`; the SDK sent `model: "deepseek-v4-flash"` and `output_config.effort: "max"`.

The synthetic user prompt was:

> 用一句简短中文解释天空为什么是蓝色，不要调用工具。

| SDK / bundled CLI | Title setting | Main requests | Title requests | Title returned by the provider and persisted by the SDK |
| --- | --- | ---: | ---: | --- |
| 0.3.220 / 2.1.220 | Default | 1 | 1 | 解释天空为什么是蓝色 |
| 0.3.232 / 2.1.232 | Default | 1 | 1 | 解释天空为什么是蓝色 |
| 0.3.257 / 2.1.257 | Default | 1 | 1 | 天空为什么是蓝色 |
| 0.3.257 / 2.1.257 | `CLAUDE_CODE_DISABLE_TERMINAL_TITLE=1` | 1 | 0 | No generated title |

All **seven real model requests returned HTTP 200**. The title requests had a `json_schema` output format with a `title` property and carried `cc_entrypoint=sdk-ts`. In all three enabled cases, `getSessionInfo()` subsequently returned the generated text in both `customTitle` and `summary`. The live tests used this third-party provider; I have not repeated them against Anthropic-hosted Claude models.

**Why the extra request occurs in ACP**

In the bundled CLI code for all three SDK versions, the input-loop title latch is initialized from conditions equivalent to the following, with names expanded for readability:

```ts
let titleRequested =
  initialMessages.some(message => message.type !== "system") ||
  isEssentialTrafficMode() ||
  disableTerminalTitle;
```

It is not unconditionally set just because the process is headless or SDK-driven. A fresh streaming session can enter the native automatic-generation branch on an eligible user message.

Separately, this PR's [turn-end handler](https://github.com/agentclientprotocol/claude-agent-acp/blob/d334766ef95dd89201979d42252e3d2a5a259cb9/src/session-titles.ts#L133-L162) reads the stored metadata and, if no custom title is visible yet, calls the [explicit generator](https://github.com/agentclientprotocol/claude-agent-acp/blob/d334766ef95dd89201979d42252e3d2a5a259cb9/src/session-titles.ts#L192-L221) with the trailing user/assistant text.

The CLI's explicit `generate_session_title` handler invokes generation without rechecking the automatic-title opt-out. Consequently:

- With native automatic generation disabled, the adapter sees no title and issues an explicit request anyway.
- With generation enabled, the adapter can race native generation and issue a duplicate request.

The adapter tests used ACP `0.75.1` at [`3e23c5b`](https://github.com/agentclientprotocol/claude-agent-acp/commit/3e23c5b960b66a6d2c892e7524c952e731c076a7) as the baseline, plus local fix candidates, with the real ACP transport/SDK/CLI and a local mock Messages API:

| Adapter behavior | Main requests | Title requests | Explicit adapter title calls |
| --- | ---: | ---: | ---: |
| Generation retained, default settings, one captured race (see note below) | 1 | 2 | 1 |
| Original adapter, native title opt-out set to `1` | 1 | 1 | 1 |
| Local candidate removing adapter-side generation, default | 1 | 1 | 0 |
| Same candidate, native title opt-out set to `1` | 1 | 0 | 0 |
| Same candidate, native title opt-out set to `0` | 1 | 1 | 0 |

The two-request capture came from a local opt-out-guard candidate with the opt-out unset, so the explicit generation path remained active. The duplicate is timing-dependent: an unmodified default run produced only the native title request. Adding a check for the existing native opt-out stopped the unwanted explicit call when that flag was set, but retained the competing generation paths when enabled. An ACP-only opt-out tested locally suppressed the explicit request but left native generation active. Setting `CLAUDE_CODE_DISABLE_NONESSENTIAL_TRAFFIC=1` alone also did not eliminate title requests in the tested build.

**Generation and publication timing are separate**

The live `0.3.257` run on September 8 showed:

```text
06:14:03.186 UTC  native title request started
06:14:03.200 UTC  main request started
06:14:07.515 UTC  main response completed
06:14:08.497 UTC  title response completed
                 SDK metadata subsequently contained the generated title
```

This means seeing the opening prompt in `summary` at turn-end does not necessarily mean no title request was made. The generated title may simply not be persisted yet. This timing difference is reproducible; I cannot tell whether it explains the original observation behind this PR.

**Possible follow-up and its tradeoffs**

I tried a local candidate that removes the adapter's text collection, undeclared-method capability check, explicit generation, and retry machinery, while retaining stored-title publication, summary fallback, deduplication, and reset handling. This lets the SDK own generation and its opt-out.

There are observable tradeoffs, so I would not describe the deletion as behavior-preserving:

- In a deliberately delayed-title mock test, a title completing after the first idle was only published to the client at the next turn's idle. Two normal turns still produced just one title request. Immediate late-title publication would need its own synchronization solution.
- The native requests observed here use the opening user message, whereas the adapter can use assistant text and raw slash-command arguments too. These experiments do not establish equivalent title coverage or quality for every session mode.

For that local candidate, two new no-generation regressions failed before the deletion and all seven focused title tests passed afterward. `npm run build`, `npm run check` (ESLint and Prettier), and `git diff --check` passed. I did not run the full test suite. The live tests above establish SDK-only behavior; the patched adapter was exercised against the mock API.

<details>
<summary>SDK-only live reproducer and configuration</summary>

The following is the SDK side of the live harness. When run directly with a real API key, it uses the configured provider without the recording proxy. It keeps the streaming session open while waiting for stored metadata, rather than closing it as soon as the main answer finishes.

Save as `sdk-title-probe.mjs` in a fresh test directory and install one of the three SDK versions there. The commands below match the macOS arm64 test environment (Node `26.8.1`); adjust the native package path for another platform.

```sh
npm install --ignore-scripts @anthropic-ai/claude-agent-sdk@0.3.257

export ANTHROPIC_BASE_URL=https://opencode.ai/zen/go
# Set ANTHROPIC_API_KEY locally with your own credential.
export ANTHROPIC_MODEL='deepseek-v4-flash[1m]'
export ANTHROPIC_DEFAULT_OPUS_MODEL='deepseek-v4-flash[1m]'
export ANTHROPIC_DEFAULT_SONNET_MODEL='deepseek-v4-flash[1m]'
export ANTHROPIC_DEFAULT_HAIKU_MODEL=deepseek-v4-flash
export CLAUDE_CODE_SUBAGENT_MODEL=deepseek-v4-flash
export CLAUDE_CODE_EFFORT_LEVEL=max
export DISABLE_TELEMETRY=1
export DISABLE_ERROR_REPORTING=1
export DISABLE_AUTOUPDATER=1
export CLAUDE_CODE_DISABLE_FEEDBACK_SURVEY=1
export CLAUDE_CODE_DISABLE_OFFICIAL_MARKETPLACE_AUTOINSTALL=1

export CLAUDE_CONFIG_DIR="$(mktemp -d)"
unset CLAUDE_CODE_DISABLE_TERMINAL_TITLE
node sdk-title-probe.mjs \
  "$PWD/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs" \
  "$PWD/node_modules/@anthropic-ai/claude-agent-sdk-darwin-arm64/claude" default

# Control run with a fresh configuration:
export CLAUDE_CONFIG_DIR="$(mktemp -d)"
export CLAUDE_CODE_DISABLE_TERMINAL_TITLE=1
node sdk-title-probe.mjs \
  "$PWD/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs" \
  "$PWD/node_modules/@anthropic-ai/claude-agent-sdk-darwin-arm64/claude" disabled
```

```js
// Official SDK only; no ACP adapter and no explicit title-generation call.
import assert from "node:assert/strict";
import { setTimeout as delay } from "node:timers/promises";
import { pathToFileURL } from "node:url";

const [sdkEntry, cli, state] = process.argv.slice(2);
const { query, getSessionInfo } = await import(pathToFileURL(sdkEntry));
const ready = Promise.withResolvers();
const inputClosed = Promise.withResolvers();
const mainDone = Promise.withResolvers();
const promptText = "用一句简短中文解释天空为什么是蓝色，不要调用工具。";
async function* input() {
  await ready.promise;
  yield { type: "user", message: { role: "user", content: [{ type: "text", text: promptText }] }, parent_tool_use_id: null, session_id: "" };
  await inputClosed.promise;
}
const q = query({
  prompt: input(),
  options: {
    cwd: process.cwd(),
    settingSources: [],
    tools: [],
    systemPrompt: "Answer the user's question in one short sentence.",
    model: process.env.ANTHROPIC_MODEL,
    maxTurns: 1,
    includePartialMessages: true,
    extraArgs: { "replay-user-messages": "" },
    pathToClaudeCodeExecutable: cli,
    stderr: (data) => process.stderr.write(data),
  },
});
let sessionId;
const consume = (async () => {
  for await (const message of q) {
    if (message.session_id) sessionId = message.session_id;
    if (message.type === "result") {
      console.log(JSON.stringify({ event: "result", at: new Date().toISOString(), subtype: message.subtype, isError: message.is_error, result: message.result, usage: message.usage, modelUsage: message.modelUsage, totalCostUsd: message.total_cost_usd }));
      mainDone.resolve(message);
    } else if (message.type === "system" && (message.subtype === "init" || message.subtype === "session_state_changed")) {
      console.log(JSON.stringify({ event: message.subtype, at: new Date().toISOString(), sessionId: message.session_id, state: message.state, model: message.model }));
    }
  }
})();
consume.catch(error => mainDone.reject(error));
try {
  await q.supportedCommands();
  ready.resolve();
  const result = await mainDone.promise;
  assert.equal(result.subtype, "success", JSON.stringify(result));
  assert.equal(result.is_error, false);
  let info;
  if (state === "disabled") {
    await delay(5000);
    info = await getSessionInfo(sessionId, { dir: process.cwd() });
  } else {
    const deadline = Date.now() + 60000;
    do {
      info = await getSessionInfo(sessionId, { dir: process.cwd() });
      if (info?.customTitle) break;
      await delay(250);
    } while (Date.now() < deadline);
  }
  console.log(JSON.stringify({ event: "stored_session", at: new Date().toISOString(), sessionInfo: info }));
  if (state === "disabled") assert.ok(!info?.customTitle);
  else assert.ok(info?.customTitle, "No generated title persisted within 60 seconds of the main result");
} finally {
  ready.resolve();
  inputClosed.resolve();
  q.close();
  await consume;
}
```

</details>

Would removing the adapter-side generator and addressing late-title publication separately fit the intended behavior, or was this targeting a particular setup where native generation is unavailable? Thanks for taking a look; I hope the versioned results and reproducer make that distinction easier to check.

EugeneTheDev:
Can you please summarize your message, do you imply that Claude SDK already generates the title natively under the hood, so custom title generation is not needed?

tcztzy:
Yes, that's what I meant, @EugeneTheDev. The SDK generated titles natively in my tests, so the extra generation call appears unnecessary. Please let me know if I've misunderstood anything. Thank you!

PR #987, opened 2026-08-12
test: account for root permission mode fallback

kocaemre:
## Summary
- update `resolvePermissionMode` tests to account for the existing root fallback for `bypassPermissions`
- add explicit coverage for the root/non-sandbox fallback path and its log message

## Verification
- `npm run build`
- `npm run check`
- `npm run test:run -- src/tests/resolve-permission-mode.test.ts`
- `npm run test:run`

### DIFF
--- a/src/tests/resolve-permission-mode.test.ts
+++ b/src/tests/resolve-permission-mode.test.ts
@@ -1,6 +1,9 @@
 import { describe, it, expect, vi } from "vitest";
 import { resolvePermissionMode, type Logger } from "../acp-agent.js";
 
+const CAN_USE_BYPASS_PERMISSIONS = process.getuid?.() !== 0 || !!process.env.IS_SANDBOX;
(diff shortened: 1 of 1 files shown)


kocaemre:
Closing this as superseded by current main. Upstream now has deterministic coverage for the same root/bypass fallback path via the injected `allowBypass` parameter in `resolvePermissionMode`, including `resolvePermissionMode("bypassPermissions", logger, false)` clamping to `default` and logging. I briefly refreshed the old branch and re-ran the focused local checks before noticing the upstream coverage: `npm run test:run -- src/tests/resolve-permission-mode.test.ts` (9 passed), `npm run build`, `npm run check`, and `git diff --check` passed on the refreshed stale-base branch. Since the remaining PR diff is now redundant and GitHub reports it conflicting, closing is cleaner than asking for review.

PR #992, opened 2026-08-13
feat: align typed session failures with AIR protocol

nikita-ashihmin:
## Summary

- align `_meta.jetbrains.air.sessionFailure` with AIR's transcript-record protocol: `id`, `revision`, broad `category`, explicit `severity`, `title`, optional `details`, and `actions`
- forward Claude's actual user-facing assistant/result text as `title`; use adapter-authored fallback text only when Claude provides no message (transport failure, worker EOF, or an internal invariant)
- reduce categories to `connection`, `access`, `limit`, `request`, `service`, and `unknown`
- limit suggested actions to `retry`, `login`, and `new_session`
- update one transcript record through `id + revision`, keep resolved entries in history, and remove clear/tombstone updates
- suppress duplicate assistant failure prose only for clients that negotiated typed failures; preserve legacy ACP behavior otherwise
- restore synthetic usage-limit failures from session history with their exact stored Claude text
- document negotiation, delivery carriers, identity/revision behavior, mappings, recovery, and compatibility in the extension spec

## Validation

- `npm run check`
- `npm run build`
- `npm run test:run` — 777 passed, 26 skipped
- `git diff --check`

## Tracking

[IJAI-993](https://youtrack.jetbrains.com/issue/IJAI-993)


src/acp-agent.ts
```diff
@@ -957,9 +962,38 @@ type AirSessionFailureCategory =
...
+  revisions: Map<string, number>;
+  active: Map<string, PublishedSessionFailure>;
+};
+
+function createSessionFailureState(): SessionFailureState {
+  return {
+    epoch: randomUUID(),
+    revisions: new Map(),
+    active: new Map(),
+  };
```
Rizzen: Please do not inflate acp-agent.ts any further, move your new logic, types etc to separate files

Rizzen (changes requested):

src/acp-agent.ts
nikita-ashihmin: Addressed in 98c055d: moved the AIR capability contract, session-failure types/mapping, replay helpers, and lifecycle controller into src/session-failure-extension.ts. acp-agent.ts now only imports and orchestrates the extension (409 lines removed from that file). Lint, targeted Prettier check, build, full tests, and diff-check pass.

src/session-failure-extension.ts
```diff
@@ -0,0 +1,439 @@
...
+      return "real_model_success";
+    default:
+      return turnId ? "next_attempt" : "real_model_success";
+  }
+}
+
+/** Owns the AIR failure lifecycle for both the live consumer and history replay.
+ *  Published revisions become active only after delivery succeeds; recovery only
+ *  removes internal active state because transcript records are historical. */
+export function createSessionFailureController(options: {
```
EugeneTheDev: Looks like it can be a normal class

nikita-ashihmin: Done — replaced the factory with a `SessionFailureController` class and updated both live and replay call sites. I also fixed the stale `OnUserDialog` test options (`requestId` is now required by the SDK); lint, format, build, and all tests pass.

PR #995, opened 2026-08-13
feat: add _session/stop_tasks to stop background sub-agents

### LINKED ISSUE #994: Client stop cancels the turn but leaves background sub-agents running — expose a stop primitive
### What happens
A client's stop cancels the current turn (`session/cancel`), but background sub-agents dispatched in that turn keep running — invisible and unkillable from the client. (This is the inverse concern to #976: there the worry is a cancel *killing* sub-agents that should survive; here it's that a client has *no way* to deliberately stop them when it wants to.)

### Expected
A client can request that background sub-agents be stopped and learn which were stopped and which could not be found.

### Actual
The SDK exposes `stopTask(taskId)` (emits a `task_notification` status `stopped`), but the adapter surfaces no stop primitive over ACP. Once dispatched, a background sub-agent can't be ended by the client that started it.

### Proposal (sketch — mirrors the existing `_session/steering` ext-method)
- Method `_session/stop_tasks`, params `{ sessionId, toolUseIds?: string[] }` (parent tool-use ids), result `{ stopped: string[], notFound: string[] }`.
- No `toolUseIds` → stop every live `isSubagent` task; background Bash spared.
- Per-task `stopTask` failures degrade to `notFound`, never fail the batch.
- Capability advertised at `initialize` (`_meta.stopTasks.supported: true`) so clients feature-detect.

### Evidence it's implementable
I've run this as a patch on the built adapter (`dist/acp-agent.js`): a probe issued the RPC, the adapter returned `{stopped:[…], notFound:[]}` with the dispatch still open, the task's card closed via a `stopped → failed` terminal, and the sub-agent terminated early (settled ~55s against a 90s sleep, no result streamed). A PR porting this to `src/acp-agent.ts` + `src/tests` follows this issue — happy to adjust the method name / capability key / result shape in review.

jstrunk:
Closes #994

## What

Adds a request extension method that lets a client stop the background tasks (sub-agents) a session has dispatched. Mirrors the existing `_session/steering` ext-method exactly (underscore-prefixed method const + param validator, handler on the agent class, `.onRequest` registration beside steering, capability advertised in the `initialize` `_meta`).

```
_session/stop_tasks  { sessionId: string, toolUseIds?: string[] }
                  ->  { stopped: string[], notFound: string[] }
```
(both arrays are parent `tool_use` ids — the card ids a client already renders.)

## Why

`session/cancel` aborts the turn, but the SDK's background tasks survive it by design. In the settled-dispatch shape there is no turn left to cancel at all, so a client's stop button is a full no-op while the sub-agents run on invisibly and unkillably. There was previously no ACP surface to stop them (no `stopTask` reference in the adapter).

## Semantics

- With `toolUseIds`: stop the live tasks whose spawning Agent/Task `tool_use` id matches.
- Without `toolUseIds`: stop every live task marked `isSubagent`. Background Bash jobs are deliberately spared — they are not `isSubagent`, matching the discriminator used across the `liveBackgroundTasks` registry (a shell can outlive every turn; its contract is a wake-on-exit notification, not a turn-scoped stop).
- Each stop is `session.query.stopTask(taskId)`, awaited so the response reports what the SDK actually accepted. A per-task failure (the task settled between snapshot and stop, or the SDK rejects an unknown id) degrades that id to `notFound` rather than failing the batch.
- Reading `liveBackgroundTasks` is safe here: a stoppable task has not settled, so its entry is live by definition. A task that finishes in the snapshot→stop window simply lands in `notFound` — the honest answer.
- Capability advertised at `initialize` via `_meta.stopTasks.supported: true`, mirroring `_meta.steering.supported`. This lets a client gate the call; an unpatched agent answering `-32601` would otherwise be silently swallowed by fire-and-forget senders.

## Downstream

`stopTask` causes the SDK to emit a `task_notification` with status `'stopped'`, which flows through the existing `task_notification` handling — no new terminal path is introduced here.

## Testing

`npm run build` (tsc) is clean; `npm run lint` and `prettier --check` are clean. New unit tests in `src/tests/acp-agent.test.ts` under `describe("stop background tasks (_session/stop_tasks)")`, all passing:

- `stops the tasks named by explicit toolUseIds and returns them in `stopped``
- `reports an unmatched explicit toolUseId in `notFound``
- `without toolUseIds stops every live sub-agent and spares background Bash`
- `a stopTask that throws lands the id in `notFound` and does not fail the batch`
- `advertises the stopTasks capability at initialize`
- `rejects when the session is unknown`
- `rejects when the query stream has already closed`

Full suite: 771 passed, 2 failed — both failures are pre-existing on clean `main` (context-window-seeding assertions in `create-session-options.test.ts`), unrelated to this change and reproduced with these changes stashed.

Behavioral probing against a running adapter (in a downstream integration) showed the expected early settle + tool card close + no result streamed for the stopped sub-agent. This does not claim OS process-level kill verification — only the observable ACP-level outcome.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -282,6 +282,54 @@ function parseSteerRequest(params: unknown): SteerRequest {
   };
 }
 
+/** Custom (extension) request method a client uses to stop the background
+ *  tasks (sub-agents) a session has dispatched. `session/cancel` aborts the
+ *  turn, but the SDK's background tasks survive it by design — and in the
+ *  settled-dispatch shape there is no turn to cancel at all, so a client's stop
+ *  button is otherwise a no-op while the sub-agents run on invisibly. Same
+ *  ext-method conventions as {@link STEER_METHOD} above; advertised to clients
+ *  via `InitializeResponse._meta.stopTasks.supported`. */
+const STOP_TASKS_METHOD = "_session/stop_tasks";
+
+/** Params of a {@link STOP_TASKS_METHOD} request. With `toolUseIds`, stop the
+ *  live tasks whose spawning Agent/Task tool_use id matches (the card ids a
+ *  client already renders); without it, stop every live `isSubagent` task.
+ *  Background Bash jobs are deliberately spared in the latter case, matching
+ *  the sub-agent discriminator the rest of the registry uses. */
+export type StopTasksRequest = {
+  sessionId: string;
+  toolUseIds?: string[];
+};
+
+/** Result of a {@link STOP_TASKS_METHOD} request, in parent tool_use ids: the
+ *  calls whose task the SDK accepted a stop for (`stopped`) and, when explicit
+ *  `toolUseIds` were requested, those that matched no live task (`notFound`).
+ *  A per-task `stopTask()` failure degrades to `notFound` rather than failing
+ *  the batch. */
+export type StopTasksResponse = {
+  stopped: string[];
+  notFound: string[];
+};
+
+/** Validate raw JSON-RPC params into a {@link StopTasksRequest}. `sessionId` is
+ *  required; `toolUseIds`, if present, must be an array of non-empty strings. */
+function parseStopTasksRequest(params: unknown): StopTasksRequest {
+  if (!params || typeof params !== "object") {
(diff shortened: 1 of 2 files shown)


benbrandt:
superseded by #1017

PR #1002, opened 2026-08-14
feat: switch providers for loaded Claude sessions

serjic:
### Problem

Provider routing is captured when the Claude SDK `Query` is created. As a result, `providers/set` and `providers/disable` updated newly created sessions but an already loaded ACP session continued using its previous backend.

User and project Claude settings could also override the ACP-selected URL or credentials after the subprocess environment was assembled, making the configured proxy ineffective.

### Solution

- Serialize provider changes and wait for submitted turns to settle.
- Close and recreate every loaded SDK query with the same Claude session ID, cwd, MCP configuration, and ACP creation parameters.
- Block new/load/resume/prompt requests only while a provider update is in progress.
- Apply the active ACP route in both the subprocess environment and programmatic settings tier.
- Clear competing Anthropic, Bedrock, Vertex, OAuth, API-key, and `apiKeyHelper` routing while an ACP override is active; preserve unrelated settings.
- Restore native/environment routing when the provider is disabled.
- Add opt-in routing diagnostics through `CLAUDE_AGENT_LOGS` without logging request content or secret headers.

### Tests

Added coverage for:

- native behavior when Providers API is unused;
- live recreation of loaded sessions;
- waiting for active turns;
- multiple loaded sessions across native → proxy A → proxy B → native;
- user settings precedence and preservation of unrelated settings;
- Anthropic, Bedrock, Vertex, legacy gateway auth, and logout behavior.

Verification:

- `npm test`: 789 passed, 20 skipped
- `npm run build`
- `npm run check`
- `git diff --check`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -3,6 +3,8 @@
 import { resolveSettings } from "@anthropic-ai/claude-agent-sdk";
 import { claudeCliPath, runAcp } from "./acp-agent.js";
 import packageJson from "../package.json" with { type: "json" };
+import { appendFileSync, mkdirSync } from "node:fs";
+import { join } from "node:path";
 
 // `--cli` is checked first so that `--version`/`-v` (and any other flags) are
 // forwarded to the wrapped native CLI rather than swallowed by our own version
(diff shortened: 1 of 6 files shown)


src/index.ts
```diff
@@ -61,7 +63,39 @@ if (process.argv.includes("--cli")) {
...
+        const logFile = join(logDirectory, "agent.log");
+        const writeRoute = (...args: unknown[]) => {
+          const rendered = args
+            .map((arg) => (arg instanceof Error ? (arg.stack ?? arg.message) : String(arg)))
+            .join(" ");
+          appendFileSync(logFile, `${new Date().toISOString()} pid=${process.pid} ${rendered}\n`);
+        };
+        return {
+          log: (...args: unknown[]) => {
+            console.error(...args);
```
EugeneTheDev, line 80: Do we always want to log any messages with `log` as errors in the console? Seems too strict

src/index.ts
```diff
@@ -61,7 +63,39 @@ if (process.argv.includes("--cli")) {
...
+            const first = args[0];
+            if (
+              typeof first === "string" &&
+              (first === "Claude ACP started" ||
```
EugeneTheDev: This match on hardcoded strings looks fragile, why do we need it?

src/index.ts
```diff
@@ -61,7 +63,39 @@ if (process.argv.includes("--cli")) {
...
+                first.startsWith("[providers/") ||
+                first.startsWith("[session/query]") ||
+                first.startsWith("Waiting for ") ||
+                first.startsWith("Recreating Claude session "))
+            ) {
+              writeRoute(...args);
+            }
+          },
+          error: (...args: unknown[]) => console.error(...args),
```
EugeneTheDev: Currently `error` logs only to the console. Don't we want to log to the file too?

EugeneTheDev (changes requested):
In general looks good, got a few minor questions

EugeneTheDev (approved):

PR #1004, opened 2026-08-14
feat: align Claude modes and clear-context planning

nikita-ashihmin:
## Summary
- translate Claude SDK permission suggestions and tool-specific decisions into ACP v1 presentation and editable option metadata
- align Claude session permission modes and clear-context plan handling
- split permission effects, modes, normalization, option construction, presentation, and response handling into focused modules
- document the permission extension and add unit plus live-audit coverage for native edge cases

## Tests
- `npm run check`
- `npm run build`
- `env -u ANTHROPIC_BASE_URL npm run test:run` (931 passed, 27 skipped)

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -79,7 +79,6 @@ import {
   SDKMessage,
   SDKMessageOrigin,
   SDKPartialAssistantMessage,
-  SDKResultMessage,
   SDKUserMessage,
   SlashCommand,
   ThinkingConfig,
@@ -173,7 +172,13 @@ import {
   toolUpdateFromToolResult,
 } from "./tools.js";
 import { nodeToWebReadable, nodeToWebWritable, Pushable, unreachable } from "./utils.js";
-import { continuePlanInFreshContext } from "./clear-context-coordinator.js";
+import {
+  acceptedPlanToolResult,
+  containsToolResultFor,
+  ExitPlanCoordinator,
+  executionDiagnostic,
+  exitPlanModeRawOutput,
+} from "./exit-plan.js";
 import { DEFAULT_AGENT_ID, EFFORT_CONFIG_ID } from "./session-config-ids.js";
 
 export { DEFAULT_AGENT_ID, EFFORT_CONFIG_ID } from "./session-config-ids.js";
@@ -196,24 +201,6 @@ function sanitizeTitle(text: string): string {
     return sanitized;
   }
   return sanitized.slice(0, MAX_TITLE_LENGTH - 1) + "…";
-}
-
-function acceptedPlanToolResult(
-  notification: SessionNotification,
-  toolUseId: string | undefined,
-): SessionNotification {
-  const update = notification.update;
-  if (
-    !toolUseId ||
-    update.sessionUpdate !== "tool_call_update" ||
-    update.toolCallId !== toolUseId
-  ) {
-    return notification;
-  }
-  const completed = { ...update };
-  delete completed.rawOutput;
-  delete completed.content;
-  return { ...notification, update: { ...completed, status: "completed" } };
 }
 
 /**
@@ -1405,10 +1392,7 @@ export class ClaudeAcpAgent {
   providerConfig?: ProviderConfig;
   /** Serializes provider changes while every open query is recreated between turns. */
   private providerUpdate: Promise<void> | null = null;
-  /** Abort controllers for clear-context session replacements. Cancellation or
-   * teardown invalidates the lease so an async createSession cannot resurrect
-   * a closed public session or attach its turn to a newer replacement. */
-  private clearContextRestarts = new Map<string, AbortController>();
+  private readonly exitPlan = new ExitPlanCoordinator<Session, Turn>();
   /** Grace period before a `session/cancel` forces a wedged prompt loop to
    *  return "cancelled". See {@link DEFAULT_FORCE_CANCEL_GRACE_MS}. Mutable so
    *  tests can shrink it. */
@@ -4680,7 +4664,7 @@ export class ClaudeAcpAgent {
   }
 
   async cancel(params: CancelNotification): Promise<void> {
-    this.clearContextRestarts.get(params.sessionId)?.abort();
+    this.exitPlan.cancel(params.sessionId);
     const session = this.sessions[params.sessionId];
     if (!session) {
       return;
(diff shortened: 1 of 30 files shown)


or1can:
Data point in favour of this direction, from a self-hosted stdio MCP server whose write tools carry `_meta["anthropic/requiresUserInteraction"]: true`.

On 0.68.0 the prompt for those tools shows **Deny / Allow Once / Always Allow**, and choosing Always Allow does nothing observable: no rule is written to `.claude/settings.local.json`, to user settings, or to `allowedTools` in `~/.claude.json`, and the next call prompts identically. That follows from the current `updatedPermissions: suggestions ?? [{type: "addRules", rules: [{toolName}], behavior: "allow", destination: "session"}]` — the synthesised session rule is then ignored by the CLI, which enforces the annotation regardless. So the gate holds and the button is inert, but a prompt that offers to remember an answer and silently doesn't is how people learn to click through prompts without reading.

Both halves of this PR's summary look like they fix it: "expose at most one durable option backed by the exact Claude `suggestions` change set" and "remove synthesized broad rules".

Worth knowing while implementing that, from patching `sdk.mjs` to log the raw `can_use_tool` control requests (tool names sanitised):

```json
{ "tool_name": "mcp__example__write_tool", "requires_user_interaction": true, "tool_use_id": "toolu_…" }
{ "tool_name": "Bash", "permission_suggestions": [ { "type": "addRules", "…": "…" } ], "tool_use_id": "toolu_…" }
```

An annotated tool arrives with **no** `permission_suggestions` at all, so keying the durable option off the suggestions set is sufficient here — no need to wait on the flag.

The flag itself is on the wire but is dropped by `Query.processControlRequest` before it reaches `canUseTool`, so hosts cannot currently read it: anthropics/claude-agent-sdk-typescript#415. Once that lands, honouring it explicitly would also cover any future case where the CLI sends suggestions *and* requires interaction.

Versions: `@agentclientprotocol/claude-agent-acp` 0.68.0, `@anthropic-ai/claude-agent-sdk` 0.3.232 (CLI 2.1.232), Zed on macOS.

nikita-ashihmin:
Thanks for the concrete report — it caught a regression in the rewritten branch. Fixed in 48624e6.

MCP tools now expose a durable option only when the SDK supplies `permission_suggestions`, and selecting it returns that exact provider payload. Without suggestions — including the current `requires_user_interaction` path — the prompt offers only one-time allow and reject. I also added regression coverage for regular and Computer Use MCP tools.

Once anthropics/claude-agent-sdk-typescript#415 exposes the flag to `canUseTool`, we can additionally honor it explicitly.

src/permissions/options/shell.ts
```diff
@@ -0,0 +1,115 @@
...
+}
+
+function displayList(values: string[]): string {
+  if (values.join(", ").length > 50) return "similar";
+  if (values.length === 1) return values[0]!;
+  if (values.length === 2) return `${values[0]} and ${values[1]}`;
+  return `${values.slice(0, -1).join(", ")}, and ${values.at(-1)}`;
+}
+
+function displayPaths(paths: string[]): string {
```
EugeneTheDev: This functions is very dense which makes it harder to read and understand. Consider adjusting the formatting to introduce new lines to visually split function body into logical blocks. Also consider adding concise comments for the function itself and for key parts of implementation, again, to make it easier to understand

src/permissions/modes.ts
```diff
@@ -0,0 +1,57 @@
...
+
+interface PermissionModeLogger {
+  error: (...args: unknown[]) => void;
+}
+
+// Bypass Permissions doesn't work if we are a root/sudo user.
+const IS_ROOT = (process.geteuid?.() ?? process.getuid?.()) === 0;
+export const ALLOW_BYPASS = !IS_ROOT || !!process.env.IS_SANDBOX;
+
+const PERMISSION_MODE_ALIASES: Record<string, PermissionMode> = {
```
EugeneTheDev: Add a short doc comments explaining why these aliases are needed

src/acp-agent.ts
```diff
@@ -1499,6 +1356,31 @@ class ClientConnection implements AcpClient {
...
+  return new Promise<T>((resolve, reject) => {
+    const cleanup = () => signal.removeEventListener("abort", onAbort);
+    const onAbort = () => {
+      cleanup();
+      reject(new Error("Tool use aborted"));
+    };
+    signal.addEventListener("abort", onAbort, { once: true });
+    if (signal.aborted) {
+      onAbort();
+      return;
```
EugeneTheDev: **Bug: Unhandled promise rejection when signal is already aborted**

When `signal.aborted` is true at call time, `onAbort()` is called and then `return` exits the executor — `void operation.then(...)` is never reached. The `operation` promise (already created by the caller before `raceWithAbort` was invoked — JS evaluates arguments before the call) now has no rejection handler. If `client.requestPermission` or `client.sessionUpdate` rejects asynchronously after the signal aborted, Node 22 treats it as an `UnhandledPromiseRejection` and crashes by default.

The fix is to remove the `return;` on this line. Without it, `void operation.then(...)` always attaches handlers. If the outer promise is already settled (rejected by `onAbort`), the inner `resolve`/`reject` calls are silent no-ops, so there are no side effects and the rejection is consumed.

src/acp-agent.ts
```diff
@@ -5491,181 +5479,101 @@ export class ClaudeAcpAgent {
...
-      if (session.modes.currentModeId === "bypassPermissions" && !matchedAskRule) {
-        return {
-          behavior: "allow",
-          updatedInput: toolInput,
-          updatedPermissions: suggestions ?? [
-            { type: "addRules", rules: [{ toolName }], behavior: "allow", destination: "session" },
-          ],
+      if (parentToolUseId) {
+        presentation.toolCall._meta = {
+          claudeCode: { parentToolUseId },
```
EugeneTheDev: **Bug: `toolName` missing from `_meta.claudeCode` on permission-request tool calls from subagent tools**

The `ToolUpdateMeta` type (line 953-955) declares `claudeCode.toolName: string` as a required field. The existing comment reads: *claudeCode metas always carry `toolName`, so clients can rely on one shape everywhere.*

When `parentToolUseId` is set, this line writes `{ claudeCode: { parentToolUseId } }` — `toolName` is absent. `buildClaudePermissionPresentation` does not set `_meta.claudeCode` on the `toolCall`, so there is nothing to merge from.

Compare `ensureToolCallEmitted` (line 5381-5388), which correctly merges `{ ...(update._meta?.claudeCode || {}), parentToolUseId }` preserving `toolName`. The old code used `satisfies ToolUpdateMeta`, which would have caught the omission at compile time.

Fix: `claudeCode: { toolName, parentToolUseId }` (both fields).

src/clear-context-coordinator.ts
```diff
@@ -0,0 +1,160 @@
...
+    }
+  };
+  assertRestartActive();
+  const turn = oldSession.activeTurn;
+  if (!turn || turn.settled || host.currentSession(sessionId) !== oldSession) {
+    throw new Error("Cannot clear context without an active ACP turn");
+  }
+
+  const params = restartParams(oldSession);
+  turn.carriedUsage = { ...oldSession.accumulatedUsage };
```
EugeneTheDev: **Plausible: non-abort `restartSession` failure is misclassified as `transport_lost` to the client**

Three mutations happen before `await host.restartSession(...)` with no rollback on failure:
1. `turn.carriedUsage` is set (line 127)
2. `oldSession.pendingExitPlanContextReset` is cleared (line 128) — prevents any retry of the clear-context flow
3. `host.closeQueryStream(oldSession)` closes the stream (line 129)

If `restartSession` throws a non-abort error (network timeout, auth failure, SDK error), `ExitPlanCoordinator.restart()` rethrows without cleanup (the catch in `exit-plan.ts` only cleans up on abort). The error reaches the outer catch in `acp-agent.ts`'s message loop, which classifies it as a transport failure and fails the turn as `transport_lost` — masking the real cause from the client.

src/permissions/normalization.ts
```diff
@@ -0,0 +1,130 @@
...
+function isDestination(value: unknown): value is PermissionUpdateDestination {
+  return typeof value === "string" && DESTINATIONS.has(value as PermissionUpdateDestination);
+}
+
+function safePlainString(value: unknown, maxLength: number): value is string {
+  if (typeof value !== "string") return false;
+  const normalized = value.trim();
+  return (
+    normalized.length > 0 &&
+    normalized.length <= maxLength &&
```
EugeneTheDev: **Bug: length cap checked on trimmed string, but original untrimmed value is stored**

`safePlainString` computes `normalized = value.trim()` (line 46) and checks `normalized.length <= maxLength` here, but returns `value is string` asserting the original (untrimmed) value is valid. A tool name like 510 spaces followed by `"Bash"` (514 chars total) passes `MAX_TOOL_NAME_LENGTH = 512` because trimmed `"Bash"` is 4 chars, and gets stored as a 514-char string with leading spaces.

Permission rules stored with whitespace-padded tool names silently never match real tool invocations, making the allow/deny rule ineffective. Fix: apply the length check to the original value, not the trimmed copy — or trim before storing.

src/permissions/options/shell.ts
EugeneTheDev, line 18: Dense function without comments. Consider adding doc comments and concise explanatory comments for key parts of the implementation as well. 

Also, consider adding new lines to split the implementation in a few logical chunks and make it easier to read through

src/permissions/modes.ts
EugeneTheDev, line 11: Consider adding a comment explaining, why these aliases are needed

src/acp-agent.ts
```diff
@@ -1520,6 +1403,33 @@ export class ClaudeAcpAgent {
     this.sessions = {};
     this.client = client;
     this.logger = logger ?? console;
+    this.exitPlan = new ExitPlanCoordinator<Session, Turn>({
+      currentSession: (id) => this.sessions[id],
```
EugeneTheDev, line 1406: Why these lambdas are not just regular functions in the `ExitPlanCoordinator` class? This should separate the concerns and make it more readable

EugeneTheDev (changes requested):

src/acp-agent.ts
nikita-ashihmin: Fixed in 79d3d4a.  now always attaches fulfillment and rejection handlers to the already-created operation, even when the signal is already aborted. Added a regression test for a client operation that rejects after aborting the signal during request creation.

src/clear-context-coordinator.ts
nikita-ashihmin: Fixed in 79d3d4a. A non-abort replacement failure is now settled by  as the original turn-scoped error instead of escaping into the query consumer's  catch. The reset/usage mutations are delayed until a replacement exists, cleanup is centralized, and a coordinator regression test covers the failure path.

src/permissions/normalization.ts
nikita-ashihmin: Fixed in 79d3d4a. The cap is now checked against the original string length while the trimmed value is still used only for the non-empty/control-character validation. Added coverage for a 512-space prefix followed by .

src/permissions/options/shell.ts
nikita-ashihmin: Addressed in 79d3d4a. Added a doc comment describing the path-shortening contract and concise comments around the suffix disambiguation and native label summarization steps, with logical spacing between them.

src/permissions/modes.ts
nikita-ashihmin: Addressed in 79d3d4a. Added a comment explaining that the aliases bridge case-insensitive user-facing settings (including Claude Code legacy shorthand) to the SDK's camel-cased wire values.

src/acp-agent.ts
nikita-ashihmin: The protocol-specific restart workflow and lifecycle are now owned by  (including state publication, continuation-message construction, cancellation, replacement cleanup, and failure settlement). The remaining constructor lambdas are intentionally the host-adapter boundary for concrete  state and private operations such as the session map, query teardown, audit completion, and turn settlement; moving those into the generic coordinator would couple it back to the full agent. The new non-abort failure handling in 79d3d4a is also implemented inside the coordinator.

EugeneTheDev (approved):

PR #1020, opened 2026-08-21
Fix/replay persisted subagents

euri10:
## Summary

Fix `session/load` replay losing tool calls made inside persisted Claude
subagents.

## Problem

The ACP [`session/load` contract](https://agentclientprotocol.com/protocol/v1/session-setup#loading-a-session)
requires the agent to replay the entire conversation through `session/update`
notifications. Tool executions are represented by
[`tool_call` updates](https://agentclientprotocol.com/protocol/v1/tool-calls#creating).

Claude stores subagent messages separately from the main session transcript:

- `getSessionMessages()` returns the main transcript.
- `listSubagents()` discovers persisted subagents.
- `getSubagentMessages()` returns each subagent transcript.

`replaySessionHistory()` only loaded the main transcript. As a result, a nested
tool call appeared during the original live turn but silently disappeared when
the same session was loaded. The parent Agent/Task result remained visible,
making the replay look complete while omitting the subagent's actual work.

This affected the adapter's
[nested subagent transcript behavior](https://github.com/agentclientprotocol/claude-agent-acp#nested-subagent-transcripts).
The client capability controls whether nested updates are rendered or flattened;
it does not transfer session-history recovery to the client.

## Fix

- Discover persisted subagents with `listSubagents()`.
- Load their histories with `getSubagentMessages()`.
- Recursively interleave each subagent history after its launching Agent/Task
  tool call.
- Reuse the existing replay conversion so capability filtering and
  `parentToolUseId` attribution remain unchanged.

## Regression test

The commits intentionally preserve red/green ordering:

1. `fece865` adds a split-storage regression fixture matching the Claude SDK's
   persisted-session layout.
2. Before the implementation, the test fails:

   ```text
   Expected: ["parent-agent-call", "child-tool"]
   Received: ["parent-agent-call"]
   ```

3. `e8f2df4` adds the implementation and makes the same test pass.

## Real-session ACP evidence

The original live session emitted the nested call with its parent relationship:

```json
{
  "method": "session/update",
  "params": {
    "sessionId": "<redacted>",
    "update": {
      "sessionUpdate": "tool_call",
      "toolCallId": "toolu_01AB1x1SdnkvcFoPF6EcmTqr",
      "rawInput": {
        "command": "ls -d <project>/*/ | xargs -n1 basename | paste -sd, -"
      },
      "_meta": {
        "claudeCode": {
          "parentToolUseId": "toolu_018NnXLu4qGQc7e4bei8C52a"
        }
      }
    }
  }
}
```

Before this fix, loading the same session emitted no corresponding child
`tool_call` update. After this fix, the child update is replayed with its
parent attribution and appears in the exported transcript in the expected
parent → child → assistant order.

Full logs are not included because they contain session content and local
environment details. The deterministic regression test provides reproducible
coverage; this excerpt documents the real ACP behavior that motivated it.

## Verification

- Focused regression test: passed
- Full Vitest suite: 811 passed, 20 skipped
- ESLint: passed
- Prettier: passed
- TypeScript build: passed
- Manual `session/load` verification against a real persisted Claude session:
  nested child tool call restored with the correct parent ID

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -63,6 +63,8 @@ import {
   FastModeDisabledReason,
   FastModeState,
   getSessionMessages,
+  getSubagentMessages,
+  listSubagents,
   listSessions,
   McpServerConfig,
   ModelInfo,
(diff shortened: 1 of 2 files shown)


euri10:
Superseded by the newer native-subagent replay implementation now on main. Rebasing this PR onto current main leaves no applicable source changes; preserving the old head as archive/fix-replay-persisted-subagents-c159150.

euri10:
Closing as superseded by current main.

PR #1026, opened 2026-08-22
Fix replay of forked session history

interkelstar:
## Summary
- persist fork lineage separately from Claude session transcripts
- replay inherited history for full and point forks, including nested forks
- resume an empty nested fork from its nearest persisted source session
- remove a fork lineage record with its session

Claude keeps a new fork’s subsequent messages under the new session ID but does not materialize the inherited transcript there. Loading that ID therefore produced an empty ACP transcript even though the model retained the parent context. A nested fork of a newborn fork also cannot resume that unborn Claude session ID, so the adapter now follows lineage to the nearest resumable ancestor and preserves the inherited branch point.

## Tests
- `npm run check`
- `npm run build`
- `npm run test:run -- src/tests/fork-history.test.ts src/tests/acp-agent.test.ts`

`npm test` has one pre-existing failing context-window test on the clean `v0.70.0` tag as well: `session/load seeds the window from the resumed session’s getContextUsage report` expects `888000` but receives `200000`.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -150,7 +150,12 @@ import {
   isFileChangeAuditTool,
   supportsAgentFileChangeReport,
 } from "./file-change-audit.js";
-import { ForkHistoryStore, forkedSessionHistory, forkResumeTarget } from "./fork-history.js";
+import {
+  ForkHistoryStore,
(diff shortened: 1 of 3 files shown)


carrutstick:
I think this is obviated by #1046

PR #1029, opened 2026-08-23
feat: report async task readiness and CLI provenance

nathansmithopenclaw-alt:
This PR has been rewritten and force-pushed. It originally landed a
`backgroundTasks` extension with three parts: an init capability, a snapshot of
live background tasks, and a readiness/provenance signal. #1017 has since merged
and covers the first two properly, so those are gone. What is left is the third
part, rebased onto current `main` and layered on `AsyncTaskRuntime` rather than
sitting beside it.

### The gap

#1017 gives clients the async task lifecycle. Nothing tells them whether that
lifecycle can carry anything.

Whether any task event is emitted at all depends on the Claude binary the
adapter spawned, and the client is never told which one that is.
`claudeCliPath()` resolves either an operator-configured
`CLAUDE_CODE_EXECUTABLE` or the platform binary bundled with the SDK. A binary
that does not emit task lifecycle produces an empty stream, and an empty stream
is exactly what a session with no background work produces. Two meanings, one
channel, and no way for a client to tell them apart. A client that surfaces
background work has to choose between claiming "nothing is running" when it does
not know that, and never trusting the stream at all.

### What this adds

Two facts, reported at the altitudes they actually live at.

**CLI provenance is connection-scoped.** Which branch `claudeCliPath()` takes is
fixed for the adapter process, so `initialize` reports it once:

```jsonc
"_meta": {
  "asyncTaskReadiness": { "cli": { "source": "bundled" } }  // or "configured"
}
```

**Readiness is session-scoped and monotone.** It starts `unconfirmed` and
latches to `confirmed` the first time the adapter observes a task lifecycle
event for that session, published as a single `session_info_update`:

```jsonc
"_meta": { "asyncTaskReadiness": { "readiness": "confirmed" } }
```

At most one such frame per session, however much background work it runs.

### How it composes with #1017

The latch is worth sending because the adapter observes strictly more lifecycle
than the client does. `AsyncTaskRuntime.taskStarted()` marks subagent tasks
`ignored` and publishes nothing for them, so a session whose background work is
entirely subagents proves the lifecycle is live to the adapter while the
client's async task stream stays empty for the session's whole lifetime. A
client cannot derive `confirmed` for itself in that case; this frame is the only
way it learns that its silence is real. There is a test for exactly that case,
with a non-subagent control beside it so the empty-stream assertion cannot pass
against a disabled runtime.

### Deliberate omissions

- **No new capability name.** #1017 already advertises `asyncTasks`, and the
  presence of the `asyncTaskReadiness` object is self-describing. Advertising it
  again would only duplicate what `initialize` already says.
- **The resolved CLI path is not on the wire.** It is an operator's filesystem
  layout, and nothing else in this adapter puts it in a protocol frame.
  `source` is the part a client cannot determine for itself.
- **No CLI version.** It is knowable without spawning the binary only for the
  bundled case. A field that is accurate for one source and a guess for the
  other is worse than its absence.
- **`_meta.asyncTaskReadiness` is a top-level key**, a sibling of `steering` and
  `goal`, not a member of the `air` object. Peers match that object as a closed
  shape, including this repo's own tests, so adding a key inside it would be a
  breaking change for anyone doing the same.
- **The readiness frame is gated on the client having negotiated `asyncTasks`.**
  Readiness describes that stream, so a client which opted out of it gains
  nothing from being told the stream is live. Provenance at `initialize` is
  unconditional, matching the other advertisements there.

### Compatibility

Purely additive for a client that already speaks #1017's shape. It reads
`asyncTaskReadiness` or it does not; nothing existing changes shape, and no
existing field gains or loses a value.

The diff is purely additive: it changes no existing line, in source or in
tests.

One behavioural note worth stating plainly. A client that negotiated
`asyncTasks` and runs background work now sees one additional
`session_info_update` at the head of the sequence a `task_started` produces. It
is published ahead of the subagent and async task runtimes deliberately, so it
leads that sequence rather than displacing the tail a consumer reads for a
turn's outcome. `session_info_update` carrying only a `_meta` payload is already
how the `goal` and session-failure extensions report state, so the shape is
established rather than new, but a client that enumerates a turn's updates
exactly will see the extra frame. Clients that did not negotiate `asyncTasks`
see no change at all.

### Tests

`npm run test:run`, plus `npm run lint` and `npm run format:check`. The new
coverage is:

- CLI provenance at `initialize` for configured, unset, and empty
  `CLAUDE_CODE_EXECUTABLE` (empty falls through to the bundled binary, matching
  `claudeCliPath()`'s own truthiness test), that the resolved path never appears
  in the response, and that the `air` capability object is unchanged.
- The readiness frame leads the sequence a `task_started` produces. Note the
  ordering is only *observable* on the non-subagent path, where the async task
  runtime also publishes; `AsyncTaskRuntime` ignores subagent tasks, so nothing
  else is emitted in that window and position cannot be detected there. The
  test names say which case each one actually pins rather than implying a
  guarantee the fixture cannot see.
- A subagent-only session confirms readiness while publishing no `async_task_*`
  update, with a non-subagent control proving the runtime was enabled.
- The latch holds across a second `task_started`, with a control proving both
  tasks reached the runtime.
- A session with no background work, and a client that did not negotiate
  `asyncTasks`, each publish no readiness frame, both with controls proving the
  turn actually ran.
- Fail-closed parsing of the payload, so a garbled frame can never read as proof
  that the lifecycle is live.

Each absence assertion is paired with a control in the same test proving the
turn actually ran, so "no frame was published" cannot pass on a path that never
executed.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -112,6 +112,11 @@ import {
 } from "./native-subagents.js";
 import { AIR_ASYNC_TASKS_CAPABILITY, withAirMeta } from "./air-extension.js";
 import {
+  asyncTaskCliSource,
+  type AsyncTaskReadiness,
+  type AsyncTaskReadinessMeta,
+} from "./async-task-readiness.js";
+import {
   AsyncTaskRuntime,
   backgroundBashTaskFromToolResult,
   clientSupportsAsyncTasks,
@@ -603,6 +608,10 @@ export type Session = {
   /** Last goal snapshot sent to the ACP client, used to roll back an
    *  optimistic `/goal` update when the command itself fails. */
   lastPublishedGoal?: GoalSnapshot | null;
+  /** Async task readiness for this session. Absent means `unconfirmed`; it
+   *  latches to `confirmed` on the first observed task lifecycle event and
+   *  never moves again, so the client is told exactly once. */
+  asyncTaskReadiness?: AsyncTaskReadiness;
   /** Count of result messages the consumer should treat as orphans and skip
    *  (not promote/attribute to the current head). When cancel() settles+removes
    *  a queued turn, that turn's user message was already pushed to the SDK, so
@@ -1960,6 +1969,14 @@ export class ClaudeAcpAgent {
           controlMethod: GOAL_CONTROL_METHOD,
           actions: [...GOAL_ACTIONS],
         } satisfies GoalCapability,
+        // Which binary `claudeCliPath()` resolves through is fixed for the
+        // adapter process, so async-task CLI provenance is reported once here
+        // rather than repeated on every session. A sibling of `steering` and
+        // `goal` rather than a member of the `air` object, which peers match
+        // as a closed shape.
+        asyncTaskReadiness: {
+          cli: { source: asyncTaskCliSource() },
+        } satisfies AsyncTaskReadinessMeta,
       },
     };
   }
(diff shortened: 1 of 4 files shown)


nathansmithopenclaw-alt:
Force-pushed a rewrite of this branch, and retitled it.

#1017 landed after I opened this and covers two of its three parts properly:
the init capability duplicated `asyncTasks`, and the task snapshot was a strict
subset of what `AsyncTaskRuntime` already streams. Both are gone. Rebasing the
original as written would also have regressed two subagent attribution tests,
since its snapshot competed with `subagent_state_update` for the tail of a turn.

What is left is the third part, the piece #1017 does not cover: a client is
never told which Claude binary the adapter spawned, so it cannot tell an empty
async task stream apart from a binary that never reports one. This now layers on
`AsyncTaskRuntime` instead of sitting beside it — CLI provenance once at
`initialize`, and a per-session readiness latch that confirms on the first
observed task lifecycle event. About 50 lines of implementation, rebased onto
current `main`, description updated above.

Worth flagging one thing for review: `AsyncTaskRuntime` ignores subagent tasks,
so a session whose background work is all subagents leaves the client's async
task stream empty even though the lifecycle demonstrably works. That case is why
the latch is a frame rather than something a client derives for itself, and
there is a test for it.

This is still awaiting first-time-contributor workflow approval, so CI has never
run on it. Happy to adjust the shape if the `_meta` placement or the extra
`session_info_update` is not what you want.

PR #1034, opened 2026-08-24
fix: allow raw model ids in config option

### LINKED ISSUE #1033: setSessionConfigOption rejects valid Anthropic model IDs not in supportedModels() picker
(Note: the issue below is written primarily by AI, but I reviewed it. I stand by the main description of the problem, am far less certain of the root cause / suggested fix)

## Problem

`setSessionConfigOption` for the model config option rejects any model ID that doesn't appear in the SDK's `supportedModels()` picker list, even when the ID is a valid Anthropic model that the API accepts.

For example, `claude-opus-4-8` is a valid model ID (Anthropic's API accepts it), but since `supportedModels()` has rotated to show `claude-opus-5` as the successor, attempts to set `claude-opus-4-8` via `setSessionConfigOption` fail with:

```
Invalid value for config option model: claude-opus-4-8
```

This makes it impossible for ACP clients to use valid-but-unlisted models (retired from the picker, custom deployments, newly launched models not yet in the SDK's curated list, etc.) through the standard config option mechanism.

## Root Cause

In `setSessionConfigOption` (acp-agent.ts ~line 5014-5029), the model validation tries:
1. Exact match against the picker's `allValues`
2. Round-trip safety for `currentValue`
3. `resolveModelPreference()` against `session.modelInfos`
4. `resolveModelPreference()` against option-derived infos

If all four fail, it throws `"Invalid value for config option model: <value>"`.

The `resolveModelPreference` function uses `modelVersionsCompatible()` as a guard, which extracts a version from the preference (e.g. "4.8" from "claude-opus-4-8") and rejects candidates with a different version. Since the picker no longer has any "4.8" entries, no candidate passes.

## Suggested Fix

When all resolution tiers fail for a model config option, accept the raw value as a passthrough instead of throwing:

```typescript
if (!validValue) {
  if (params.configId === MODEL_CONFIG_ID) {
    // Accept any model ID — let the Anthropic API validate it downstream.
    validValue = { value: params.value, name: params.value };
  } else {
    throw new Error(`Invalid value for config option ${params.configId}: ${params.value}`);
  }
}
```

This lets `query.setModel()` receive the raw ID and the API itself becomes the validator. Invalid IDs will surface as API errors on the next turn rather than being rejected at the picker layer.

## Impact

Any ACP client (Nori, Zed, etc.) that allows users to type custom model IDs is currently broken for models not in the SDK's curated picker. The "custom model input" feature becomes useless because the adapter rejects the value before it reaches the API.

kocaemre:
## Summary

- allows the model config option to pass through a non-empty raw model ID when it is not present in the picker
- keeps alias/current-value resolution first, so picker entries still canonicalize to the SDK option value
- covers the regression with an out-of-picker `claude-opus-4-8` model ID

Closes #1033.

## Verification

- `npm run test:run -- src/tests/session-config-options.test.ts -t 'passes through out-of-picker model IDs'`
- `npm run test:run -- src/tests/session-config-options.test.ts src/tests/model-resolution.test.ts`
- `npm run build`
- `npm run check`
- `git diff --check`

I also ran `npm run test:run`. It completed with 963 passing tests, 27 skipped, and the same two root-environment `resolve-permission-mode` failures seen when running as root: `bypassPermissions` is intentionally downgraded to `default` in that environment. The changed model config tests passed.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -5463,6 +5463,18 @@ export class ClaudeAcpAgent {
       }
     }
 
+    if (!validValue && params.configId === MODEL_CONFIG_ID) {
+      const rawModelId = params.value.trim();
+      if (rawModelId) {
+        // The picker is a convenience list, not the complete validation
+        // boundary for Anthropic model IDs. Accept a non-empty raw model id and
+        // let the SDK/API validate it on the next turn so valid-but-unlisted
+        // IDs (new launches, retired picker entries, custom gateways) remain
+        // reachable through the standard config option path.
+        validValue = { value: rawModelId, name: rawModelId };
+      }
+    }
+
     if (!validValue) {
       throw new Error(`Invalid value for config option ${params.configId}: ${params.value}`);
     }
--- a/src/tests/session-config-options.test.ts
+++ b/src/tests/session-config-options.test.ts
@@ -324,14 +324,17 @@ describe("session config options", () => {
       expect(modelOption?.currentValue).toBe("claude-sonnet-4-6");
     });
 
-    it("throws for completely invalid model value", async () => {
-      await expect(
-        agent.setSessionConfigOption({
-          sessionId: SESSION_ID,
-          configId: "model",
-          value: "gpt-4",
-        }),
-      ).rejects.toThrow("Invalid value for config option model: gpt-4");
+    it("passes through out-of-picker model IDs to the SDK", async () => {
+      const response = await agent.setSessionConfigOption({
+        sessionId: SESSION_ID,
+        configId: "model",
+        value: "claude-opus-4-8",
+      });
+
+      expect(setModelSpy).toHaveBeenCalledWith("claude-opus-4-8");
+      expect(response.configOptions.find((o) => o.id === "model")?.currentValue).toBe(
+        "claude-opus-4-8",
+      );
     });
 
     it("returns full configOptions in the response", async () => {


kocaemre:
Rebased this branch onto current `main` (`c3ff343`) and resolved the conflict in `src/acp-agent.ts` against the newer `SessionModeManager.reconcileForModel(...)` extraction.

Conflict resolution kept this PR's raw model-id pass-through behavior while using the current mode-reconciliation helper from `main`. Unknown/out-of-picker model IDs still keep the current mode instead of being spuriously downgraded.

Re-ran local checks after the rebase:

```bash
npm ci
npm run test:run -- src/tests/session-config-options.test.ts -t 'passes through out-of-picker model IDs'
npm run test:run -- src/tests/session-config-options.test.ts src/tests/model-resolution.test.ts src/tests/session-mode.test.ts
npm run build
npm run check
git diff --check upstream/main..HEAD
```

Results: targeted regression passed (`1 passed`), focused config/model/session-mode suite passed (`76 passed, 2 skipped`), build passed, lint/prettier passed, and diff check was clean.

I also reran the full suite:

```bash
npm run test:run
```

Result: `1042 passed, 27 skipped, 2 failed`. The two failures are the existing root-environment `resolve-permission-mode` expectations where `bypassPermissions` is downgraded to `default` when tests run as root; they are unrelated to this model-config change.

PR #1053, opened 2026-08-29
fix: match WSL session list cwd casing

### LINKED ISSUE #1041: session/list matches cwd case-sensitively, so a session created with a /mnt/<drive> path cannot be found with that same path
## Summary

`session/list` filters by `cwd` using the caller's exact string, while sessions appear to be indexed under a lowercased path. On WSL, where `/mnt/c/...` paths normally carry capitals, this makes the agent inconsistent with itself: create a session with a given `cwd`, then list with that very same string, and the session is not in the result.

## Environment

- `@agentclientprotocol/claude-agent-acp` 0.70.0
- Claude Code CLI 2.1.245, subscription login (`apiType=native`)
- WSL2, Ubuntu 22.04, Node 22.23.2
- cwd on a DrvFs mount, e.g. `/mnt/c/Users/Me/Notes/MyVault`

## Reproduction

Drive the agent over stdin, keeping stdin open long enough for each reply:

1. `initialize`
2. `session/new` → `{"cwd": "/mnt/c/Users/Me/Notes/MyVault", "mcpServers": []}` — succeeds, returns a session id
3. `session/list` → `{"cwd": "/mnt/c/Users/Me/Notes/MyVault", "cursor": null}` — **0 sessions**
4. `session/list` → `{"cwd": "/mnt/c/users/me/notes/myvault", "cursor": null}` — **returns the session from step 2**

Only the casing differs between steps 3 and 4.

## Observed

I have not traced the implementation. The index appears to hold the lowercased form: the project directory created for this working directory is `~/.claude/projects/-mnt-c-users-me-notes-myvault`, all lowercase, while the path passed in was `/mnt/c/Users/Me/Notes/MyVault`.

Sessions under `/home` are unaffected, being lowercase already — the symptom only shows on `/mnt/<drive>` paths.

WSL itself is not the cause: `getcwd`, `realpath` and Python's `os.path.realpath` all return `/mnt/c/Users/...` with the capitals intact, so the lowercasing happens above the filesystem.

## Suggested fix

Compare `cwd` the same way it is stored, treating `/mnt/<drive>/...` as case-insensitive the way native Windows paths already are. The invariant worth restoring is that a session created with a given `cwd` is findable with that same `cwd`.

## Related

`@agentclientprotocol/codex-acp` has the same bug, filed as agentclientprotocol/codex-acp#431. There the cause is visible in the source: a `normalizePathForComparison` helper lowercases only paths matching `^[A-Za-z]:/` or a UNC prefix, so `/mnt/c/...` takes a case-preserving branch while the stored value has no capitals.

## Downstream impact

The Obsidian Agent Client plugin passes the vault path to `session/list` for its "Show current vault only" filter. Under WSL mode that filter is always empty for any vault on a Windows drive.

kocaemre:
## Summary
- normalize `/mnt/<drive>/...` cwd filters before calling the Claude SDK session index
- keep non-WSL paths case-preserving
- add `session/list` coverage for the WSL drive-path case

Fixes #1041

## Verification
- `npm run test:run -- src/tests/acp-agent.test.ts -t 'session/list'`
- `npm run test:run -- src/tests/acp-agent.test.ts`
- `npm run build`
- `npm run check`
- `git diff --check`

I also ran `npm run test:run`; it reached `1044 passed, 27 skipped` and failed only the two existing root-environment `resolvePermissionMode` expectations where `bypassPermissions` is downgraded to `default` when tests run as root.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -1202,6 +1202,11 @@ const SUPPORTED_PROTOCOLS: LlmProtocol[] = ["anthropic", "bedrock", "vertex"];
 const PROVIDER_ID = "main";
 const DEFAULT_ANTHROPIC_BASE_URL = "https://api.anthropic.com";
 
+function normalizeSessionListDir(dir: string | undefined): string | undefined {
+  if (dir === undefined) return undefined;
+  return /^\/mnt\/[a-z](?:\/|$)/i.test(dir) ? dir.toLowerCase() : dir;
+}
+
 /**
  * Vertex needs project + region that the standard `providers/set` payload
  * (`apiType`/`baseUrl`/`headers`) does not model, so clients pass them through
@@ -2168,7 +2173,9 @@ export class ClaudeAcpAgent {
   }
 
   async listSessions(params: ListSessionsRequest): Promise<ListSessionsResponse> {
-    const sdk_sessions = await listSessions({ dir: params.cwd ?? undefined });
+    const sdk_sessions = await listSessions({
+      dir: normalizeSessionListDir(params.cwd ?? undefined),
+    });
     const sessions = [];
 
     for (const session of sdk_sessions) {
--- a/src/tests/acp-agent.test.ts
+++ b/src/tests/acp-agent.test.ts
@@ -57,6 +57,7 @@ import {
   getSessionInfo,
   getSessionMessages,
   importSessionToStore,
+  listSessions,
   PermissionUpdate,
   query,
   SDKAssistantMessage,
@@ -85,6 +86,7 @@ vi.mock("@anthropic-ai/claude-agent-sdk", async (importOriginal) => {
     forkSession: vi.fn(),
     getSessionInfo: vi.fn(),
     importSessionToStore: vi.fn(actual.importSessionToStore),
+    listSessions: vi.fn(actual.listSessions),
     // Delegates to the real implementation so integration tests that read
     // actual transcripts keep working; unit tests override per-call with
     // `mockResolvedValueOnce`.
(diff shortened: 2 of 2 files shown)


kocaemre:
Rebased this PR onto current `main` after the branch became conflicting (`DIRTY`). The only conflict was in `src/tests/acp-agent.test.ts`; I kept the newer upstream `importSessionToStore` fork-session test setup and the PR's `listSessions` mock/session-list tests together.

Validation on the refreshed branch:
- `npm ci` — passed (248 packages installed, 0 vulnerabilities)
- `npm run test:run -- src/tests/acp-agent.test.ts -t 'session/list'` — passed (2 passed, 493 skipped)
- `npm run build` — passed (`tsc`)
- `npm run check` — passed (`eslint` + `prettier --check`)
- `git diff --check origin/main..HEAD` — passed

I also ran the full `npm run test:run`; it reached 1230 passing / 27 skipped, but failed 6 tests outside this PR's touched behavior: three timeouts in existing suites, the two known root-environment `resolvePermissionMode` expectations, and one hook timeout in `session-config-options`. The targeted `session/list`, build, lint, and format checks above are green after the rebase.

kocaemre:
Rebased this PR onto current `main` (`d870ebd2`) after upstream moved and force-pushed refreshed head `9fa0b1f`.

No code changes beyond replaying the existing WSL cwd normalization commit on top of the current branch.

Local verification in this cron environment:

```bash
git diff --check origin/main..HEAD
# passed

npm run test:run -- src/tests/acp-agent.test.ts -t 'WSL|session list cwd casing|non-WSL cwd filters'
# passed: 2 tests, 493 skipped

npm run check
# passed: eslint src --ext .ts + prettier --check .

npm run build
# passed: tsc
```

I also tried the broader file-level suite:

```bash
npm run test:run -- src/tests/acp-agent.test.ts
```

It ran 495 tests and passed 485 with 9 skipped, but timed out in the pre-existing broad-suite test `stop reason propagation > forwards background output that arrives after the turn resolves (issue #679)`. The focused WSL regression tests for this PR passed after the rebase.

DCO sign-off is preserved on the PR commit.

PR #1055, opened 2026-08-29
fix: mark informational chunks with metadata

### LINKED ISSUE #1042: informational messages (hook notices) are emitted as unmarked agent_message_chunk — indistinguishable from model output; attach _meta
## Summary

`informational` SDK messages (hook notices such as UserPromptSubmit output, and similar harness-level text) are emitted as plain `agent_message_chunk` updates with the severity folded into the text as a markdown prefix (`**Notice:** …`). On the wire they are indistinguishable from model output, so clients that treat `agent_message_chunk` as "the assistant's reply" — chat UIs, TTS pipelines, anything that persists the assistant text — inevitably mix harness notices into the model's prose.

Filing this as the complement of #1030: that issue asks for the *dropped* system subtypes to be delivered at all; this one asks for the *delivered* ones to be machine-distinguishable. Whatever channel #1030's fix ends up using would want the same marker.

**Version:** `@agentclientprotocol/claude-agent-acp` 0.70.0

## Where

`dist/acp-agent.js`, the `informational` case:

```js
case "informational": {
    // Free-form notice from the SDK (e.g. why a UserPromptSubmit/Stop
    // hook blocked continuation). Surface the text so the user sees it
    // instead of a silent stop. ACP's agent_message_chunk has no
    // severity field, so fold the level into the text for the more
    // prominent levels ('info' is transcript-only noise — leave plain).
    const text = message.level === "info"
        ? message.content
        : `**${message.level[0].toUpperCase()}${message.level.slice(1)}:** ${message.content}`;
    await sendUpdate({
        sessionId: message.session_id,
        update: {
            sessionUpdate: "agent_message_chunk",
            content: { type: "text", text },
        },
    });
    break;
}
```

The comment's rationale is sound — a hook-blocked turn must not look like a silent stop — and rendering these inline is exactly right for an editor transcript. The problem is only that the client has no way to *know* it's a notice.

## Reproduction

1. Configure a `UserPromptSubmit` hook in `~/.claude/settings.json` that prints anything to stdout (a memory-reminder hook, a context injector, …).
2. Drive the adapter from any ACP client and send a prompt.
3. The turn's `agent_message_chunk` stream begins with `**Notice:** UserPromptSubmit says: <hook stdout>` followed immediately by the model's actual reply text — one channel, no marker.

## Observed impact

We run Claude Code through this adapter as a headless runner in an agent harness (chat UI on several surfaces, sessions persisted, optionally spoken via TTS). The hook notice lands:

- in the persisted assistant message (it is now part of "what the model said" forever),
- in the spoken reply for voice conversations,
- in downstream consumers of the reply text (task logs, review verdict parsing, …).

`'info'`-level messages are worse off: they are sent as *plain* content with no prefix at all, so even a heuristic on the `**Level:**` prefix cannot see them.

The only client-side workaround today is a regex on the rendered prefix (`/^\*\*(Notice|Warning|Error):\*\* /` on a complete chunk), which is fragile (legitimate model output can start with bold-prefixed text, and the format is presentation, not contract).

## Suggested fix

Attach the structured fields to the update via `_meta`, which ACP reserves exactly for this:

```js
await sendUpdate({
    sessionId: message.session_id,
    update: {
        sessionUpdate: "agent_message_chunk",
        content: { type: "text", text },
        _meta: { "claude-agent-acp": { kind: "informational", level: message.level } },
    },
});
```

Keeping the `**Notice:**` text fold as-is preserves today's rendering in clients that ignore `_meta` (zero breakage), while clients that want to route notices elsewhere (status line, collapsed block, excluded from TTS/persistence) get a reliable signal — including for `'info'`-level messages that currently carry no visible marker at all.

kocaemre:
## Summary

- tag SDK `informational` notices with `_meta.claudeCode.kind = "informational"`
- include the notice `level` so clients can separate hook/status notices from model text
- keep the existing rendered text unchanged for clients that ignore `_meta`

Fixes #1042

## Test plan

- `npm test -- --run src/tests/acp-agent.test.ts -t informational`
- `npm run build`
- `npm run check`
- `IS_SANDBOX=1 npm run test:run`
- `git diff --check`

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -3542,6 +3542,12 @@ export class ClaudeAcpAgent {
                   update: {
                     sessionUpdate: "agent_message_chunk",
                     content: { type: "text", text },
+                    _meta: {
+                      claudeCode: {
+                        kind: "informational",
+                        level: message.level,
+                      },
+                    },
                   },
                 });
                 break;
--- a/src/tests/acp-agent.test.ts
+++ b/src/tests/acp-agent.test.ts
@@ -10046,7 +10046,7 @@ describe("assembled assistant text fallback", () => {
     });
   }
 
-  it("does not re-emit the result text after an informational notice delivered it", async () => {
+  it("marks informational notices with metadata", async () => {
     const { agent, updates } = createMockAgentWithCapture();
     // A hook-blocked prompt: the SDK surfaces the block reason as an
     // informational notice and then repeats it on the result with zero output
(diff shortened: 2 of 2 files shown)


kocaemre:
Rebased this branch onto current `main` (`7c66108`, zod import fix) and re-verified the rebased head `6dd450e`.

Verification:
- `npm test -- --run src/tests/acp-agent.test.ts -t informational` — 1 passed, 461 skipped
- `npm run build` — passed
- `npm run check` — ESLint passed, Prettier check passed
- `IS_SANDBOX=1 npm run test:run` — 1045 passed, 27 skipped
- `git diff --check upstream/main..HEAD` — clean

`gh pr checks` still reports no checks on `main`, so there is no CI failure to act on after the rebase.

kocaemre:
Refreshed this PR onto current `main` (`d571358`) and resolved the informational-notice conflict by preserving the new ACP `notice` path while keeping the fallback transcript chunk tagged with `_meta.claudeCode.kind = "informational"` and the SDK `level`.

Verification on the refreshed head `9930f04`:
- `npm ci` — passed
- `npm run test:run -- src/tests/acp-agent.test.ts -t 'informational'` — passed (3 passed, 520 skipped)
- `npm run build` — passed
- `npm run check` — passed (ESLint + Prettier check)
- `git diff --check origin/main..HEAD` — clean

`gh pr checks` still reports no checks on `main`, so there is no CI failure to act on after the refresh.

PR #1059, opened 2026-08-31
fix(LLM-25012): OAuth2 Authentication for MCP Server Connections

Ololoshechkin:


src/acp-agent.ts
```diff
@@ -1193,6 +1238,159 @@ class ClientConnection implements AcpClient {
   }
 }
 
+type McpAuthenticationHost = {
```
EugeneTheDev, line 1559: Can we extract this block of code, that is not in the `ClaudeAcpAgent` class, into a separate file, to avoid inflating this already huge agent file? Smth like `mcp-auth.ts`

EugeneTheDev (approved):

src/acp-agent.ts
Ololoshechkin: Let's do in a separate PR, the file already has 19 type declarations like this :)
So better address the whole problem at once separately

PR #1063, opened 2026-09-01
fix: steered turn parks forever when the steer aborts a task-notification followup

utkmittal1:
**Symptom.** A `session/steering` message injected while the model is streaming its task-notification followup (the relay after a background subagent finishes) leaves the held `session/prompt` unresolved forever. The steered answer's content still streams, every expected SDK frame arrives, but the turn never settles; only `session/cancel` or the next prompt releases it.

**Mechanism.** The steer aborts the followup cycle, whose result is autonomous-origin and therefore banks a trailing-idle debt ("autonomous results keep their own trailers"). But the abort collapses that cycle's trailing idle into the single idle spanning the interrupted and steered cycles — so the debt lane absorbs the one idle the steering lane was waiting for, and the ready steered settle is starved. With the stream still open, no teardown rescues it.

**Fix.** In the idle handler, a ready steered settle — echo consumed, outcome recorded — takes the idle ahead of debt absorption. A not-ready steer still falls through to the debt lane, preserving the guarantee that an earlier turn's owed idle can never settle a steered cycle still running; a genuinely separate owed trailer stays banked and absorbs the next idle.

**Evidence.** Live repro (background subagent holding the turn, steer during the followup stream): stock build, prompt unresolved after 240s, three runs; patched build, resolves in ~2s, same recipe. The regression test models the exact frame sequence with the mock stream held open, so the end-of-stream teardown can't mask a missed settle. Full suite: 1045 passing.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -3322,6 +3322,30 @@ export class ClaudeAcpAgent {
                       session.owedTrailingIdles--;
                     }
                     settleDeferredIfDrained();
+                  } else if (
+                    isSteering(session.activeTurn) &&
+                    session.activeTurn.steeredEchoes.size === 0 &&
+                    session.activeTurn.steeredSettle !== undefined
+                  ) {
+                    // A READY steered settle — echo consumed, outcome
+                    // recorded — takes this idle ahead of the debt lane
+                    // below. When the steer aborted an autonomous followup
+                    // mid-write, that followup's result banked a trailer
+                    // (the "keep their own trailers" rule), but the abort
+                    // collapsed its trailing idle into this one spanning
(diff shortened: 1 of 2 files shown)


benbrandt:
Thanks @utkmittal1, this pinned down the collapsed-trailer case exactly. We're closing it in favor of #1166, which keeps your ready-first ordering with one correction. Here "ready" also counted an outcome recorded *before* the steered echo (from the cycle the steer interrupted, or a held turn's outcome), so an owed idle landing between the echo and the steered result settled the turn early. #1166 only counts a result recorded after the last echo, and adds a regression test for that ordering alongside yours.

PR #1070, opened 2026-09-02
fix: show shell commands in permission prompts

### LINKED ISSUE #1068: Bash permission prompts show the model-written description instead of the command (regression in 0.71.0)
### Summary

Since 0.71.0, `session/request_permission` for `Bash`/`PowerShell` sets the tool-call
title to the model-authored `input.description` rather than `input.command`. The command
being approved is not rendered anywhere in the prompt, so the user approves
"List files in current directory" without seeing `ls -la ~/.config/zed`.

Observed in the Zed agent panel (registry-managed `claude-acp`) on 0.73.0.

### Root cause

`src/permissions/presentation.ts:80-84`:

```ts
const shellTitle =
  value.toolName === "Bash" || value.toolName === "PowerShell"
    ? (compactText(value.input.description) ?? value.toolName)
    : undefined;
const toolCallTitle = shellTitle ?? subjectTitle ?? info.title;
```

`toolCallTitle` feeds both `toolCall.title` and `_meta.permission.title`, so both
carry the description.

The comment immediately above says the intent is to "reuse the exact standard
tool-call heading as the permission heading so the approval never maintains a
second, divergent name for the operation" — but `toolInfoFromToolUse` in
`src/tools.ts` uses `input.command` as the Bash title, so this branch creates
exactly the divergence the comment rules out. Within one transcript, an
already-approved Bash call shows the command while the pending approval for the
same tool does not.

`rawInput` still carries the command, so this is presentation-only. But there's no
path to it in the prompt: the Bash content block is either a `terminal` (empty
before approval) or the description again.

hwisu:
## Summary
- use the canonical tool-call title for permission prompts, so shell approvals show the exact command instead of the model-authored description
- map `PowerShell` through the same shell presentation path as `Bash`
- complete the PowerShell terminal lifecycle with matching `terminal_info`, `terminal_output`, and `terminal_exit` metadata, including failed commands
- add focused regression coverage without changing the existing Bash lifecycle tests

Fixes #1068

## PowerShell input and result contract
`PowerShell` is a distinct tool name, but its documented input fields match `Bash`: the executable command is in `command`, while `description` is an optional human-readable summary. The Claude Code hooks reference documents the shared `command`, `description`, `timeout`, and `run_in_background` fields, and the tools reference likewise states that PowerShell hooks receive the command in `tool_input.command`.

Anthropic closed [claude-code#83647](https://github.com/anthropics/claude-code/issues/83647#issuecomment-5310033500) after documenting this equivalence. Reusing the SDK's exported `BashInput` shape for the common input fields is therefore intentional; the SDK currently does not export a separate `PowerShellInput` type.

PowerShell results do not need a PowerShell-specific output assumption here. The result mapper first shape-checks structured stdout/stderr and otherwise falls back to the standard raw `tool_result` string or content-block array. Routing PowerShell through that existing mapper preserves this fallback while completing the adapter's capability-gated terminal sequence:

1. initial `tool_call` with `terminal_info`
2. `tool_call_update` with `terminal_output`
3. final `tool_call_update` with `terminal_exit`

All three events use the original tool-use ID, so the client can reconcile and close the terminal. Failed PowerShell results also bypass the generic error renderer and finish the terminal with a non-zero display exit code.

The permission presentation introduced in #1004 already handled `Bash` and `PowerShell` together when choosing a description-based permission title, but the canonical tool presentation and terminal result paths only handled `Bash`. This PR completes both paths.

References:
- [Claude Code hooks: PowerShell input](https://code.claude.com/docs/en/hooks#powershell)
- [Claude Code tools reference: PowerShell tool](https://code.claude.com/docs/en/tools-reference#powershell-tool)
- [Anthropic schema confirmation](https://github.com/anthropics/claude-code/issues/83647#issuecomment-5310033500)

## Verification
- `npm run check`
- `npm run build`
- `npm run test:run -- src/tests/tools.test.ts -t "PowerShell terminal output"` (1 passed)
- full suite with an empty temporary Claude config: `CLAUDE_CONFIG_DIR=<empty temporary directory> npm run test:run` (1070 passed, 27 skipped)

### DIFF
--- a/docs/permission-extension.md
+++ b/docs/permission-extension.md
@@ -81,6 +81,8 @@ The permission request carries the same standard ACP tool information used for n
 
 Compact text removes control characters, collapses whitespace where appropriate, and enforces length
 limits. Invalid optional presentation text is omitted instead of being truncated into misleading UI.
+Shell permission titles (`Bash` and `PowerShell`) preserve the full command verbatim, including
+whitespace and line breaks, without applying compact-text normalization or length limits.
 
 Permission options are fixed. When a durable suggestion contains a command prefix, path, host, or
 other rule, the adapter includes that value directly in `PermissionOption.name`, for example
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -9356,8 +9356,8 @@ function resolveSkillPath(skillName: string, cwd?: string): string | undefined {
  *  notification for a tool_use. Shared by every site that surfaces a tool call:
  *  the streamed tool_use path (first encounter → tool_call, later encounter →
  *  refine) and the permission flow (`ensureToolCallEmitted`), so they can't
- *  drift. The initial `tool_call` carries `status: "pending"` and, for Bash, the
- *  `terminal_info` _meta that the later `terminal_output`/`terminal_exit`
+ *  drift. The initial `tool_call` carries `status: "pending"` and, for shell tools,
+ *  the `terminal_info` _meta that the later `terminal_output`/`terminal_exit`
  *  updates key off of; a refining `tool_call_update` carries neither. */
 function toolCallNotification(
   toolUse: { id: string; name: string; input: unknown },
@@ -9378,7 +9378,7 @@ function toolCallNotification(
   return {
     _meta: {
       claudeCode: claudeCodeMetaFromToolUse(toolUse, cwd),
-      ...(toolUse.name === "Bash" && supportsTerminalOutput
+      ...((toolUse.name === "Bash" || toolUse.name === "PowerShell") && supportsTerminalOutput
         ? { terminal_info: { terminal_id: toolUse.id } }
         : {}),
     } satisfies ToolUpdateMeta,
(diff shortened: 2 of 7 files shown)


biphobe:
It would be great to get this merged, claude acp is pretty much unusable due to commands not being shown when asked for approval.

levrik:
@biphobe I downgraded to 0.70.0 until this fix gets merged.

rmecham:
I would characterize this as a critical bug fix. Any plans to merge?

rmecham:
> @biphobe I downgraded to 0.70.0 until this fix gets merged.

How does one do that?

levrik:
@rmecham Add this to you Zed config under the `agent_servers` key:

```json
"claude-acp-0.70.0": {
  "type": "custom",
  "command": "npx",
  "args": ["-y", "@agentclientprotocol/claude-agent-acp@0.70.0"]
},
```

It will take some time when selecting it for the first time (as `npx` needs to pull stuff) but then it just works.

PR #1071, opened 2026-09-02
fix(auth): keep the JSON-RPC auth_required rejection for capable clients

### LINKED ISSUE #1072: An api_retry with authentication_failed publishes an access warning that outlives the auth_required rejection
## What happens

When a running CLI process loses its credentials (for example, the user runs `claude auth logout` in a terminal), the next prompt goes through this sequence:

| Time | SDK message | Agent output |
|---|---|---|
| +0.2 s | `system/api_retry`, `error: "authentication_failed"`, attempt 1 of 10 | session-scoped `_meta` failure: category `access`, severity `warning`, title "Retrying Claude, attempt 1 of 10.", no actions |
| +0.7 s | synthetic login assistant message | JSON-RPC `auth_required` rejection |

The CLI gives up on the first attempt, re-reads the credentials, and reports the sign-out. The retry notice is a credential re-check, not a real retry loop. But the agent has already published it as a typed session failure, and nothing clears that row when the turn is rejected with `auth_required` a moment later. Its `recoveryPolicy` is `next_attempt`, so it survives until the next turn starts.

For a client that renders typed failures, the visible result after the user dismisses the auth prompt is a lock-icon row "Retrying Claude, attempt 1 of 10." with no action, instead of the signed-out state.

Seen with `@anthropic-ai/claude-agent-sdk` 0.3.257 (CLI 2.1.257) and `claude-agent-acp` at #1071.

## Expected

An `api_retry` whose `error` is `authentication_failed` (or `oauth_org_not_allowed`) publishes no session failure. The sign-out that follows is the real signal, and it already carries the `login` action. A rejected turn must not leave a "Retrying" row behind.

## Where

`src/acp-agent.ts`, the `api_retry` case, currently maps every non-null `error_status` through `providerFailureCategory(message.error)` and publishes a warning.

## Related

Client-side handling of the sign-out is in #1071. This issue is separate: it is about the retry warning, not the sign-out dedupe.

### LINKED ISSUE #1073: The typed session failure replaced the JSON-RPC auth_required rejection, so a capable client's auth flow never starts
## What happens

Since #979, a client that negotiates `_meta.jetbrains.air` with the `sessionFailure` capability no longer gets the JSON-RPC `auth_required` error when the CLI is signed out. The agent settles the turn as a successful `end_turn` and attaches the typed failure to the prompt response instead.

ACP defines the `auth_required` error as the signal that starts the client's own auth flow. AIR parks the refused prompt on it, shows its auth method chooser, and resumes the prompt after sign-in. With the rejection gone, the chooser never appears and the refused prompt is lost.

Two smaller defects sit on the same path:

1. **One sign-out publishes two failure rows.** The CLI reports a refused turn twice: as a synthetic assistant message ("Not logged in · Please run /login"), then as the error-shaped result with the same text. Each one published a typed failure, so the older row kept a dead `login` action.
2. **The row title is CLI TUI advice.** "Not logged in · Please run /login" has no meaning in an ACP client. The policy already has a client-neutral fallback title ("Sign in to continue using Claude.") that is not used.

## Expected

- A refused turn answers with the JSON-RPC `auth_required` error for every client, also for one that supports typed session failures.
- A capable client additionally gets one session-scoped `access` failure. Its `login` action is the way back in after the user dismisses the client's auth prompt.
- One sign-out publishes one failure while the previous sign-out is still active.
- The failure title is the client-neutral fallback. The CLI text travels as expandable `details`.

## Where

`src/acp-agent.ts`, `failActiveWithSessionFailure`: for a capable client it calls `settleActive` with an `end_turn` outcome for every kind, including `auth_required`.

## Related

#1072 covers the `api_retry` warning that the same sign-out produces. Fix for both: #1071.

### LINKED ISSUE #1074: A sign-in done outside the query process leaves the signed-out failure active, so the next sign-out publishes no login action
## What happens

The agent clears the signed-out session failure only on the SDK's `auth_status` message. That message reports a login the query process itself runs. A client can sign the user in out of band: AIR runs `claude auth login` in a terminal, in a separate process. The query process never sees it and sends no `auth_status`.

The stored sign-out record then survives the sign-in. The next prompt succeeds, and the client hides the old row after the successful turn. But the record stays active in the agent. On the next real sign-out, the `auth_required` dedupe reads that stale record and publishes nothing. The client gets the JSON-RPC rejection and opens its auth prompt, but after the user dismisses it there is no failure row with the `login` action, which is the way back in.

Seen with `@anthropic-ai/claude-agent-sdk` 0.3.257 in one session: sign-out at 17:16 published the failure, terminal login, four successful turns, sign-out at 17:21 published nothing.

## Expected

A real model answer proves the session is signed in. The success path clears the signed-out record too, so a later sign-out publishes a fresh failure with the `login` action.

## Where

`src/acp-agent.ts`, the result handler: the clear after a real model success covers `real_model_success` and warnings, but not the `auth_status` recovery policy.

## Related

#1073 restores the rejection and the session-scoped failure. #1072 removes the retry warning. Fix for all three: #1071.

fastasturtle:
## Problem

#1073: ACP defines the `auth_required` JSON-RPC error as the signal that starts the client's own auth flow. AIR parks the refused prompt, shows its method chooser, and resumes the prompt after sign-in. #979 attached the typed failure to a successful `end_turn` response for capable clients, which replaced that signal. The chooser never appeared, and the refused prompt was lost.

Also, the CLI reports one sign-out twice (the synthetic login assistant message and the error-shaped result), so one sign-out published two failure rows, and their title was CLI TUI advice ("Not logged in · Please run /login") that has no meaning in an ACP client.

#1072: a related defect. When a running CLI loses its credentials, the next request gets a 401 and the CLI emits one `api_retry` with `authentication_failed` before it reports the sign-out. The agent published that retry as an `access` warning that nothing cleared, so a capable client showed "Retrying Claude, attempt 1 of 10." with no action in place of the signed-out state.

#1074: a sign-in done outside the query process (AIR runs `claude auth login` in a terminal) sends no `auth_status`, so the stored sign-out record survives. The next sign-out reads the stale record and publishes nothing, and the `login` action is lost.

## Change

- Every client keeps the JSON-RPC `auth_required` rejection.
- A capable client additionally receives the signed-out state as one session-scoped `access` failure. Its `login` action remains the way back in after the client's auth prompt is dismissed.
- The failure title is the client-neutral policy fallback. The CLI text travels as expandable `details`.
- One sign-out publishes one failure. Only an active session-scoped `auth_required` error suppresses a repeat.
- An `api_retry` whose error maps to `auth_required` publishes no warning. The sign-out that follows is the real signal.
- A real model answer clears the stored sign-out record too, so a later sign-out publishes a fresh failure.
- The second delivery of one sign-out no longer logs "cannot fail active turn" as an error.

## Tests

- `rejects authentication_failed and publishes a session-scoped access failure`
- `does not republish auth_required when the result repeats the login text`
- `publishes no retry warning for an authentication_failed api_retry`
- `republishes the signed-out state after a real model answer cleared the last one`
- The two `auth_status` recovery tests now assert the rejection and the session update instead of the prompt response.

`npm test`, `npm run check`, and `npm run build` pass locally.

Fixes #1073.
Fixes #1072.
Fixes #1074.

Tracked in IJAI-1374.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -3050,12 +3050,36 @@ export class ClaudeAcpAgent {
 
     /** Complete a negotiated terminal failure on the prompt response itself,
      *  which is the canonical AIR carrier. Legacy clients keep the historical
-     *  JSON-RPC rejection path. */
+     *  JSON-RPC rejection path.
+     *
+     *  `auth_required` is the exception: ACP defines its JSON-RPC error as the
+     *  signal that starts the client's own auth flow (AIR parks the refused
+     *  prompt, shows its method chooser, and resumes the prompt after sign-in),
+     *  so replacing it with a successful `end_turn` would disable that flow.
+     *  Every client keeps the rejection; capable clients additionally receive
+     *  the signed-out *state* as one session-scoped failure whose `login`
+     *  action remains the way back in after the client's auth prompt is
+     *  dismissed. Its title stays the policy's client-neutral fallback — the
+     *  CLI's own text ("… Please run /login") is TUI advice, meaningless in an
+     *  ACP client, so it travels as expandable details instead. */
     const failActiveWithSessionFailure = async (
       kind: ClaudeFailureKind,
       error: unknown,
       title?: string,
     ) => {
+      if (kind === "auth_required") {
+        // One sign-out arrives twice — the synthetic login assistant message
+        // and the turn's error-shaped result repeat the same text — and the
+        // signed-out state does not change in between: publish once, and skip
+        // republishing while the previous sign-out is still active. A retry
+        // warning of the same kind is not a sign-out and must not suppress it.
+        if (!sessionFailures.hasActiveSessionError(kind)) {
+          await publishSessionFailure(kind, { turnScoped: false, details: title });
+        }
+        // Preserve legacy codes: only explicit `/login` signals trigger ACP's auth flow.
+        failActive(error);
+        return;
+      }
       if (!supportsAirSessionFailures(this.clientCapabilities)) {
         failActive(error);
         return;
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -3077,7 +3077,10 @@ export class ClaudeAcpAgent {
           await publishSessionFailure(kind, { turnScoped: false, details: title });
         }
         // Preserve legacy codes: only explicit `/login` signals trigger ACP's auth flow.
-        failActive(error);
+        // The first delivery rejects the turn, so the second one finds it settled.
+        // That is the normal course here, not the lost-failure case `failActive`
+        // logs an error for.
+        if (session.activeTurn && !session.activeTurn.settled) failActive(error);
         return;
       }
       if (!supportsAirSessionFailures(this.clientCapabilities)) {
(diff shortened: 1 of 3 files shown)


Ololoshechkin (approved):

benbrandt:
It seems that this landed in #1079

PR #1115, opened 2026-09-10
fix: bound steered turn settlement

### LINKED ISSUE #1114: Steered turn can become permanently unsettleable: terminal result absorbed by settleOrDefer, idle steer lane has no fallback
## Summary

A turn that has been steered (had a prompt injected mid-turn via `steer()`) can end up **permanently unsettleable**: its terminal result is absorbed and never delivered, `turn.resolve()` is never called, and the pending `session/prompt` request therefore never resolves. The client sees the agent as busy indefinitely.

Observed in production on `0.75.1`. The same code shape is present in `0.66.0`, and the relevant blocks are identical between the two versions, so this is not a recent regression.

Code references below are to the published `dist/acp-agent.js` (we vendor the npm package); the corresponding TypeScript sources should map directly.

## Observed behaviour

- A turn was running. Two prompts were injected into it mid-turn (each going through `steer()`).
- The turn finished normally: the underlying Claude Code CLI produced its terminal result and reported the turn over. We can date this precisely from an independent signal — the CLI's own `Stop` hook fired and successfully reported idle.
- The `session/prompt` request for that turn **did not resolve for a further ~30 minutes**. No turn-end notification of any kind reached the client in that window.
- It only recovered when a further prompt was submitted to the same session, which drained the stuck turn through the hand-off path.

So the turn did not hang without answering — **it answered, and the answer was swallowed.**

## Mechanism

Three things combine.

**1. `isSteering()` is a presence test, not an emptiness test.**

```js
function isSteering(turn) {
    return turn != null && turn.steeredEchoes !== undefined && !turn.settled;
}
```

`steer()` arms `turnInFlight.steeredEchoes` (`(turnInFlight.steeredEchoes ??= new Set()).add(steeredUuid)`). From that moment the turn reads as "steering" until it settles — an emptied `Set` still satisfies `!== undefined`.

**2. `settleOrDefer` absorbs every result for a steering turn, including the genuine terminal one.**

```js
if (isSteering(session.activeTurn)) {
    session.activeTurn.steeredSettle = outcome;
    return;
}
```

The outcome is stored, not delivered. `turn.resolve()` is not called, so the `session/prompt` RPC stays pending.

**3. The only exit is a gate with no fallback branch.**

In the idle handler:

```js
else if (isSteering(session.activeTurn)) {
    const steered = session.activeTurn;
    if (steered.steeredEchoes?.size === 0 && steered.steeredSettle !== undefined) {
        steered.deferredSettle = steered.steeredSettle;
        steered.steeredEchoes = undefined;
        steered.steeredSettle = undefined;
        settleDeferredIfDrained();
    }
}
```

If that inner condition is not met, **the idle is silently discarded** — there is no `else`, and the surrounding comment deliberately prevents the idle from reaching the no-result detector below ("swallow it, and never let it reach the #825 fail below"). That behaviour is correct for the case it was written for (an idle arriving before the steered answer), but it means a gate miss is both unrecoverable and undetectable.

The branch is also positioned below `else if (session.owedTrailingIdles > 0) { session.owedTrailingIdles--; }`, which can consume the settling idle before the steer lane is reached.

`steeredEchoes` has exactly two write sites — armed in `steer()`, cleared only inside the successful settle branch above. There is no timer on this path; the only turn-lifecycle timer is `forceCancelTimer`, which is armed exclusively inside `cancel()`.

**Net effect: a steered turn that misses that gate can neither settle nor be detected as failed.**

## Which gate held

We could not determine from the code face alone which of three conditions caused the miss in our incident, and we did not have bridge-side instrumentation running at the time. The candidates, all consistent with the observed symptoms:

1. `owedTrailingIdles > 0` consumed the settling idle. Several comments describe over-counting as "benign (absorbs one future idle)" — on a steered turn that absorbed idle may be the only settle opportunity.
2. `steeredEchoes.size > 0` — an echo never came back.
3. An arm/drain target mismatch: `steer()` arms on `session.turnQueue.find(t => !t.settled)` (its comment notes this "covers both the activated turn and one just submitted but not yet echoed/activated"), while the drain deletes from `session.activeTurn.steeredEchoes`. If those are different `Turn` objects at arm time, the uuid is never drained.

We are happy to add instrumentation and report back with the specific gate if that would be useful — please say which fields you would want logged.

## Suggested fix

**This project already solves exactly this problem on the neighbouring path.** `cancel()` arms `forceCancelTimer` with `DEFAULT_FORCE_CANCEL_GRACE_MS = 30_000`, and the comment states the rationale precisely:

> "if a turn is actively consuming the query and `interrupt()` doesn't make the SDK yield ... force the consumer to settle the active turn 'cancelled' after the floor elapses so the pending `session/prompt` still resolves per the ACP cancellation contract instead of hanging forever"

The steering path has the same hazard — a pending `session/prompt` that can hang forever — and no equivalent backstop.

Two shapes, in our order of preference:

1. **Arm an analogous backstop when `steer()` arms `steeredEchoes`**, mirroring `forceCancelTimer`, disarmed on successful settle. This covers all three candidate gates, because it does not depend on identifying which one missed.
2. **Give the steer lane an `else`** that falls through to the existing no-result handling after N unmet idles. Smaller, but only covers the gate-miss cases, not an arm/drain target mismatch.

We are glad to open a PR for either if you would prefer that to an issue.

## Versions

- Reproduced against `0.75.1`.
- The relevant blocks (`isSteering`, `settleOrDefer`'s steering early-return, the idle steer lane, the `owedTrailingIdles` ordering, and the echo-drain site) are **identical** in `0.66.0`.
- `isEmptyUserInterruptionDiagnostic`, new in the `0.7x` line, sits on the cancel-race path and does not touch any of the above.

## Why this matters to consumers

A host that keeps a turn open has no way to distinguish "still working" from "answered and swallowed": no error is raised, no notification is emitted, and the no-result detector is explicitly bypassed on this path. From the host's side the only observable is a `session/prompt` that never returns, with no bound on how long it may not return.

kocaemre:
## Summary
- Add a bounded steering settlement backstop for turns that have recorded a terminal result but are still waiting for undrained steered echoes.
- Give the steering idle lane priority over stale trailing-idle debt once an outcome exists, so the only settlement signal cannot be consumed by older debt.
- Add regression coverage for the #1114 missing-echo/stale-idle-debt hang while preserving existing pre-echo idle behavior.

Fixes #1114

## Test Plan
- `npm run test:run -- src/tests/acp-agent.test.ts -t 'steered turn|priority'` — passed (11 passed, 483 skipped)
- `npm run test:run -- src/tests/acp-agent.test.ts` — passed (485 passed, 9 skipped)
- `npm run build` — passed
- `npm run check` — passed (`eslint src --ext .ts` + `prettier --check .`)
- `git diff --check origin/main..HEAD` — passed

Note: Claude Code delegation was unavailable in this cron shell because `claude auth status --text` reports `Not logged in. Run claude auth login to authenticate.`; implementation and validation were performed directly with local tests.

### DIFF
--- a/src/acp-agent.ts
+++ b/src/acp-agent.ts
@@ -678,6 +678,10 @@ type Turn = {
   /** What a steered turn settles with once its steered work has run: the outcome
    *  of its latest result, so its usage covers every cycle the turn ran. */
   steeredSettle?: PromptResponse;
+  /** Backstop armed when a steered turn has an outcome but is waiting for an
+   *  echo that may never replay. Mirrors cancel's bounded-hang protection for
+   *  the steering lane. */
+  steeredSettleTimer?: ReturnType<typeof setTimeout>;
   carriedUsage?: AccumulatedUsage;
   /** `carriedUsage`'s per-model counterpart, so a turn that survives a
    *  clear-context restart keeps the `_meta.quota` rows it earned pre-restart. */
@@ -1144,6 +1148,13 @@ function disarmForceCancel(session: Session): void {
   if (session.forceCancelTimer) {
     clearTimeout(session.forceCancelTimer);
     session.forceCancelTimer = undefined;
+  }
+}
+
+function disarmSteeredSettle(turn: Turn): void {
+  if (turn.steeredSettleTimer) {
+    clearTimeout(turn.steeredSettleTimer);
+    delete turn.steeredSettleTimer;
   }
 }
 
@@ -3674,6 +3685,7 @@ export class ClaudeAcpAgent {
       this.finishFileChangeAudit(session, turn, auditReason);
       // Captured before the settled flip below (isHeldOpen tests !settled).
       const wasHeld = isHeldOpen(turn);
+      disarmSteeredSettle(turn);
       turn.settled = true;
       turn.usageMarkdownAbort?.abort();
       disarmForceCancel(session);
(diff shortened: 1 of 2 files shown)


kocaemre:
Refreshed this branch onto current `main` (`d571358`) and resolved the stale conflict in `src/acp-agent.ts` by preserving the newer async `settleDeferredIfDrained()` flow while keeping the bounded steered-turn backstop from this PR.

Local verification on refreshed head `a9ade83`:

- `npm ci` ✅
- `npm run test:run -- src/tests/acp-agent.test.ts -t 'settles a steered turn when its result arrives but the steered echo never drains'` ✅ (1 passed, 522 skipped)
- `npm run build` ✅
- `npm run check` ✅
- `git diff --check origin/main..HEAD` ✅

`gh pr view` now reports the PR as mergeable with only `src/acp-agent.ts` and `src/tests/acp-agent.test.ts` in the file list; this repo still reports no branch checks via `gh pr checks`.

benbrandt:
Thanks for digging into #1114, @kocaemre! We're closing this in favor of #1166. A review found problems that come from the backstop timer: nothing clears it when the steered echo replays, so it can end a turn that is still working after 30s. Moving the steer branch ahead of owed-idle debt for every steering turn also lets an earlier turn's late idle settle a steered cycle that's still running. #1166 settles on the result that names the steer (`user_message_uuids`), which we confirmed live on CLI 2.1.280, so it needs no timer.

PR #1125, opened 2026-09-12
fix: tolerate managed settings startup failures

### LINKED ISSUE #1124: Unguarded top-level `await resolveSettings()` in src/index.ts kills the agent at startup on a transient fs error
### What happens

`claude-agent-acp` exits with code 1 before it serves anything, printing only:

```
code: "EINTR"

      at ene (<home>/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs:200:52993)

Bun v1.4.0 (macOS arm64)
```

`ene` is the SDK's minified `resolveSettings`. To the ACP client this is just a
process that exited 1 with no protocol traffic, so it reaches the user as "the
agent exited with code 1".

### Why

`src/index.ts` awaits `resolveSettings` at the top level with nothing around it
([src/index.ts#L50](https://github.com/agentclientprotocol/claude-agent-acp/blob/main/src/index.ts#L50)):

```ts
const policy = await resolveSettings({ settingSources: [] });
for (const [key, value] of Object.entries(policy.effective.env ?? {})) {
  process.env[key] = value;
}
```

That is a top-level `await` in an ESM entry point, so a rejection fails module
evaluation and the process is gone. Two details make it the only call that can
do this:

- `SettingsManager.loadAllSettings()` in `src/settings.ts` wraps the same SDK
  call in `try`/`catch` and falls back to `{}`, so the per-session path cannot
  kill the process.
- The `process.on("unhandledRejection", ...)` handler is registered *after* this
  await, so it does not cover it either.

So any transient filesystem error while the managed-settings tier is read —
EINTR, EMFILE, EAGAIN — is fatal at startup.

### How we hit it

Running the adapter under Bun on macOS, with an endpoint-security client in the
`open()` path (which widens the syscall window a lot), and a dozen agents
starting at once. Bun before 1.4.1 surfaced EINTR from `node:fs` to JavaScript
instead of retrying the way Node and libuv do — fixed in
[Bun 1.4.1](https://bun.com/blog/bun-v1.4.1) — so the throw reached this
unguarded await. We saw it twice in three days across a small fleet.

Upgrading Bun removed our trigger, but not the shape of the bug. The published
bin is `#!/usr/bin/env node`, and a `node` that is really Bun is a common setup;
nothing in the package rejects it. More to the point, this call is best-effort
by nature: on a machine with no MDM or managed policy the result is empty and
the agent behaves identically. Starting with no policy env is strictly better
than not starting.

### Suggested fix

Match what `src/settings.ts` already does:

```ts
let policy;
try {
  policy = await resolveSettings({ settingSources: [] });
} catch (error) {
  console.error("Failed to resolve managed policy settings:", error);
}
for (const [key, value] of Object.entries(policy?.effective.env ?? {})) {
  process.env[key] = value;
}
```

`console.error` is safe here: stdout is only reassigned a few lines further
down, and this already goes to stderr, so the ACP stream stays clean.

### Versions

| | |
|---|---|
| `@agentclientprotocol/claude-agent-acp` | 0.72.0 — `main` is unchanged here today |
| `@anthropic-ai/claude-agent-sdk` | 0.3.257 |
| runtime | Bun 1.4.0 |
| OS | macOS 15.7.9 arm64 |

kocaemre:
## Summary
- Fixes #1124.
- Moves managed-policy env loading into a best-effort startup helper.
- Redirects diagnostics to stderr before the managed settings lookup so fallback warnings cannot corrupt ACP stdout framing.
- Adds focused coverage for both applying managed env values and continuing when `resolveSettings()` rejects.

## Test Plan
- `npm run check` — passed
- `npm run build` — passed
- `npm run test:run -- src/tests/managed-policy-env.test.ts` — passed (2 tests)
- `npm run test:run` — blocked by pre-existing/unrelated failures in this root cron environment:
  - `src/tests/resolve-permission-mode.test.ts` expects `bypassPermissions`, but running as root intentionally falls back to `default`.
  - `src/tests/session-config-options.test.ts` hook timeout in full-suite run.

Duplicate check:
- `gh pr list --repo agentclientprotocol/claude-agent-acp --state open --limit 100 ...` showed no open PR matching `resolveSettings`, transient startup, or #1124.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
-import { resolveSettings } from "@anthropic-ai/claude-agent-sdk";
 import { claudeCliPath, runAcp } from "./acp-agent.js";
+import { applyManagedPolicyEnv } from "./managed-policy-env.js";
 import packageJson from "../package.json" with { type: "json" };
 import { appendFileSync, mkdirSync } from "node:fs";
 import { join } from "node:path";
@@ -43,21 +43,19 @@ if (process.argv.includes("--cli")) {
   console.log(packageJson.version);
   process.exit(0);
 } else {
-  // Apply env vars from the managed-policy tier before any SDK call so the
-  // SDK subprocess inherits them. Going through resolveSettings (vs. a raw
-  // read of managed-settings.json) also picks up MDM sources on macOS and
-  // HKLM/HKCU on Windows.
-  const policy = await resolveSettings({ settingSources: [] });
-  for (const [key, value] of Object.entries(policy.effective.env ?? {})) {
-    process.env[key] = value;
-  }
-
-  // stdout is used to send messages to the client
-  // we redirect everything else to stderr to make sure it doesn't interfere with ACP
+  // stdout is used to send messages to the client; redirect diagnostics before
+  // managed settings are read so startup warnings cannot corrupt ACP framing.
   console.log = console.error;
   console.info = console.error;
   console.warn = console.error;
   console.debug = console.error;
+
+  // Apply env vars from the managed-policy tier before any SDK call so the
+  // SDK subprocess inherits them. Going through resolveSettings (vs. a raw
+  // read of managed-settings.json) also picks up MDM sources on macOS and
+  // HKLM/HKCU on Windows. If this best-effort lookup fails, continue startup
+  // with the existing process environment so the ACP server can still answer.
+  await applyManagedPolicyEnv();
 
   process.on("unhandledRejection", (reason, promise) => {
     console.error("Unhandled Rejection at:", promise, "reason:", reason);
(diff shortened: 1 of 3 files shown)


kocaemre:
Current `main` now includes the same managed-policy startup guard via #1146 (`58c5db1`), including extracted `applyManagedPolicyEnv` coverage and the non-rejecting failure path this PR was targeting.

I re-verified the current upstream implementation locally:

- `npm ci` ✅
- `npm run test:run -- src/tests/managed-policy.test.ts` ✅ (4 passed)
- `npm run build` ✅
- `npm run check` ✅
- `git diff --check origin/main..HEAD` ✅
- `git status --short` clean ✅

Since the patch is now superseded by current `main` and this PR is conflicting, I’m closing it to avoid keeping duplicate/dirty review queue noise. Thanks.

PR #1106, opened 2026-09-08
feat: add ACP-friendly MCP status and usage aliases

nikita-ashihmin:
## Summary

- advertise a read-only `/mcp` command to ACP clients
- render configured MCP servers compactly with status, scope, version, tools, and errors
- use the structured usage renderer for `/usage` and its Claude Code aliases `/cost` and `/stats`
- show every available quota window, including overall, OAuth app, model-specific, and extra-usage limits
- hide the adapter's internal file-change audit server from user-facing MCP output

## Testing

- `npm run build`
- `env -u ANTHROPIC_BASE_URL npm run test:run` (1239 passed, 27 skipped)
- `npx eslint src/acp-agent.ts src/usage-markdown.ts src/mcp-status-markdown.ts src/tests/acp-agent.test.ts`
- `npx prettier --check src/acp-agent.ts src/usage-markdown.ts src/mcp-status-markdown.ts src/tests/acp-agent.test.ts`

Issue #642, opened 2026-05-09
Built-in slash commands /usage /status /model /memory /permissions /agents /mcp emit no output when invoked


Seluj78:
*AI disclosure: investigation and first-draft prose written by Claude (Opus 4.7). The framing, idea, the scope cut, and this disclosure line are mine (the human opening this).*

## Description

Several built-in slash commands are advertised by the agent via `session/update` → `available_commands_update`, but produce no observable output when actually invoked through `session/prompt`. The session completes normally (no hang, unlike #435) — there is just nothing for an ACP client to render.

Affected commands observed in claude-agent-acp current release:

- `/usage`
- `/status`
- `/model`
- `/memory`
- `/permissions`
- `/agents`
- `/mcp`

By contrast `/context` *does* emit content (a markdown text body via `agent_message_chunk`) — so this isn't a blanket "no slash command output" regression.

## Steps to reproduce

1. Connect any ACP client to claude-agent-acp.
2. Confirm via `available_commands_update` that the commands above are advertised.
3. Send `session/prompt` with `text: "/usage"` (or any of the others above).
4. Observe the resulting session updates.

## Expected behavior

Either:

- **(a)** the command emits a textual equivalent of its native Claude Code TUI rendering (e.g. `/model` lists current + alternative models in markdown, `/permissions` lists current allow/deny rules), so ACP clients have something to display; or
- **(b)** the command is omitted from `available_commands_update` (since it is interactive-only and has no ACP-renderable output); or
- **(c)** `available_commands_update` gains a metadata field marking these commands as "interactive only / no ACP output", so clients can hide them from `/`-pickers instead of letting users hit empty replies.

## Actual behavior

Command appears in the slash picker, user invokes it, agent acknowledges the prompt and the turn ends with no `agent_message_chunk`, no `tool_call`, no error. Client-side this looks like a no-op — but only because the agent had nothing renderable to send for a TUI-only feature.

## Notes

These all map to interactive Claude Code TUI surfaces in the native CLI (model picker modal, permissions editor, MCP server list, etc.). It is reasonable that they have no plain-text equivalent — but advertising them in `available_commands_update` without that signal forces every ACP client to either filter them with a hardcoded blacklist or paper over the empty turn with a fallback notice.

Related (not duplicate):

- #33 — `/mcp` not recognized at all (predates `available_commands_update` rollout?). My case is the opposite: `/mcp` *is* advertised but produces no output.
[...]

PR #1136, opened 2026-09-15
feat: handle /mcp reconnect, enable and disable

kyleshay:
Part of #642 (the `/mcp` actions; `/mcp` status itself is covered by #1106).

Related: #1106 adds a read-only `/mcp` status reply in the same `prompt()` path. The two don't overlap in behaviour: this PR only takes `reconnect`, `enable` and `disable`, and leaves `/mcp` and `/mcp list` to whatever answers them. #1106 also advertises `/mcp` as "Show status; reconnect, enable, or disable MCP servers"; this PR is what makes those actions work. Happy to rebase onto #1106 if it lands first.

## Problem

In an ACP client, `/mcp reconnect <server>`, `/mcp enable <server>` and `/mcp disable <server>` answer:

> Reconnect, enable, and disable aren't available in this session.

The CLI handles these inline only in its interactive UI. The SDK already exposes the same operations as control requests (`reconnectMcpServer`, `toggleMcpServer`, `mcpServerStatus`), but the adapter never calls them.

## Change

`/mcp reconnect|enable|disable [<server>|all]` still runs through the normal SDK turn, the same way `/usage` does, so ordering, cancellation, persistence and replay are unchanged. The turn carries the parsed command. When the CLI's output for that turn is its exact headless refusal, the adapter runs the action through the SDK's control requests and publishes its reply in place of the refusal, once, whichever message shape (local command output, result, or synthetic assistant message) carries it. Any other output passes through untouched, so if the CLI starts handling these actions headless, its own output wins with no adapter change. `/mcp`, `/mcp list`, help and unknown actions are not touched.

The output replacement sits beside the existing `/usage` one (`takeUsageMarkdown`); #1106 folds `/usage` into a shared `structuredCommand`, and this would move into that too if it lands first.

The branches and replies mirror the CLI's own inline `/mcp` handler, so a user sees the same wording in an ACP client as in the terminal:

- no server name means `all`, and the action is case-insensitive
- the `ide` server is left out, and so is the adapter's own `claude_agent_acp` file-change audit server, so `/mcp disable all` cannot switch off the audit
- `reconnect all` retries only failed and needs-auth servers; a disabled or connecting server is not retried
- `all` runs in parallel and reports counts
- the "Run `/mcp enable x`" style hint is added only when the name is a plain argument (`^\w[\w.@-]*$`) and the reply stays within 1024 characters

## Replies

| Command | Situation | Reply |
|---|---|---|
| any | no servers | No MCP servers are configured. Add one with `claude mcp add`. |
| any `<x>` | unknown name | There's no MCP server named "x". Run `/mcp` in the terminal to see configured servers. |
| any | request throws | Couldn't reconnect "x" — reason. Run `/mcp` in the terminal to check. |
| reconnect x | ok | Reconnected "x". |
| reconnect x | disabled | "x" is disabled. Run `/mcp enable x` to bring it back. |
| reconnect x | connecting | "x" is already reconnecting — retries can take a few minutes when a server keeps failing. |
| reconnect x | still down | Couldn't reconnect "x" (needs authentication). Authenticate with `/mcp` in the terminal. / Couldn't reconnect "x" (not connected). Check its config with `/mcp` in the terminal. |
| reconnect all | ran | Reconnected N of M MCP server(s). Run `/mcp` in the terminal to see status. |
| reconnect all | nothing to retry | N MCP server(s) are disabled. Run `/mcp enable all` to bring them back. / All enabled MCP servers are already connected or connecting. |
| enable x | ok | Enabled "x". |
| enable x | not up yet | Enabled "x", but it isn't connected yet (connecting). Check its config with `/mcp` in the terminal. |
| enable x | already on | "x" is already enabled. / "x" is already enabled but not connected. Run `/mcp reconnect x` to retry. |
| disable x | ok | Disabled "x". |
| disable x | refused | Couldn't disable "x" — it may have been removed, or its configuration couldn't be read. Run `/mcp` in the terminal to check. |
| disable x | already off | "x" is already disabled. |
| enable/disable all | ran | Enabled N MCP server(s) (K enabled but not yet connected) (F couldn't be changed). Run `/mcp` in the terminal to see status. |
| enable/disable all | nothing to change | All MCP servers are already enabled. / All MCP servers are already enabled, but N isn't connected — reply `/mcp reconnect all` here to retry. / All MCP servers are already disabled. |

## Differences from the CLI

- The CLI's cross-session notes ("disabled in another session", "re-enabled in another session, so this disable didn't persist") depend on CLI state the SDK does not expose, so they are omitted.
- "Pending approval" is omitted: `McpServerStatus` has no such state.
- The CLI reads the resulting state from the action's return value; the SDK methods return `void`, so the adapter reads `mcpServerStatus()` afterwards.
- Recognising the refusal is an exact string match. If the CLI rewords it, the adapter falls back to forwarding the CLI's output, which is today's behaviour.
- A reloaded session replays the CLI's persisted refusal rather than the adapter's reply, as `/usage` replays the CLI's original output today.
- If `mcpServerStatus()` itself throws, the reply is the "Couldn't <action>" line above; the CLI has no equivalent case.

## Testing

- `src/tests/mcp-commands.test.ts`: parser, every reply above, and the turn path: the command is forwarded, the refusal is replaced once even when it arrives in two message shapes, non-refusal output passes through with no action run, and plain `/mcp` is untouched.
- `npm run build`, `npm run check`, `npm run test:run` pass (`providers.test.ts` has an intermittent timeout that also occurs without this change).
- Driven end to end over ACP with the real CLI against a stdio MCP server: disable, enable, reconnect, `all`, an unknown name, plus `/mcp` and an unknown action still answered by the CLI.


benbrandt:
Closing in favor of #1106

Issue #1077, opened 2026-09-02
0.73.0 ExitPlanMode: clear-context accept is unrecoverable via session/load, mode changes are never published, bypassPermissions unreachable


xintaofei:
Three problems in the ExitPlanMode permission layer as rebuilt in 0.73.0. They share the same origin (the `permissions/` rewrite + the new clear-context path from #492), so I'm filing them together — happy to split if you'd rather triage them separately.

The first one is silent data loss and is the reason I went looking; the other two I found while narrowing it down.

**Environment**

- `@agentclientprotocol/claude-agent-acp` 0.73.0
- `@anthropic-ai/claude-agent-sdk` 0.3.257 (bundled `claude`, no `CLAUDE_CODE_EXECUTABLE`)
- node v24.18.0, macOS 15 / arm64, non-root (so `ALLOW_BYPASS === true`)

**How I tested**

A minimal stdio ACP client. It advertises `fs: {readTextFile, writeTextFile}` + `terminal`, creates a session with `mcpServers: []`, sets mode to `plan`, and prompts with:

> Read notes.txt and plan one tiny change: append a line 'bye'. Produce a short plan and call ExitPlanMode immediately. Do not do anything else.

Then it answers the resulting `session/request_permission` with a chosen `optionId` and logs everything. Every observation below is from that harness, not from reading the source alone.

---

## 1. Accepting a plan with "clear context" makes the session unrecoverable via `session/load`

**Severity: silent truncation — the load succeeds and returns a conversation that is missing the entire implementation half.**

When the user picks `exit-plan-clear-auto` (or the `-bypass` / `-accept-edits` variants), `continuePlanInFreshContext` calls `host.restartSession(params, { publicSessionId: sessionId, … })`, which reaches `createSession`. There:

```js
// dist/acp-agent.js:5460
if (creationOpts?.resume === undefined || creationOpts?.forkSession) {
    options.sessionId = creationOpts.publicSessionId ? randomUUID() : sessionId;
}
```

So the public ACP session id is deliberately preserved, but Claude's private conversation is re-keyed to a **fresh random uuid**. That uuid is never communicated to the client over ACP.

The live turn is fine — it continues on the unchanged public id and ends `end_turn`, exactly as designed. The damage shows up on disk and on the next reconnect.

After one clear-context turn, `~/.claude/projects/<slug>/` holds two transcripts:

```
[...]

PR #1139, opened 2026-09-17
fix: publish effective mode after plan approval

s-cork:
## Problem

Ordinary ExitPlanMode approval changes the SDK permission mode but leaves ACP clients displaying Plan. When Auto is unavailable for the current model, clients must see the effective fallback mode.

## Change

After an allowed ExitPlanMode permission response, publish both current_mode_update and config_option_update using the session mode after applyPermissionFallback. Rejection and clear-context handling remain unchanged.

Partially addresses #1077. This does not repair session restoration after clear-context approval. The bypass-option change is independent.

## Validation

- npm run build and npm run check pass.
- Focused acp-agent and session-permission tests: 496 passed, 10 skipped.
- Regression cases cover ordinary default, bypass, Auto and Auto-to-acceptEdits fallback, including stored mode and both ACP notifications.
- Bypass test uses an existing supported bypass-only configuration; this PR does not depend on offering bypass alongside Auto.

AI assisted implementation and tests. Live Claude smoke remains pending because no authenticated Claude session is available in this environment.


benbrandt:
Thanks @s-cork! I've carried this over to #1163 with your commit kept as-is, plus a small follow-up so that a failed mode notification can't fail the approval. The fork is owned by an organization, so maintainers can't push to it. Closing in favour of #1163.

PR #1141, opened 2026-09-17
fix: show the shell command, not its description, in permission prompts

### LINKED ISSUE #1068: Bash permission prompts show the model-written description instead of the command (regression in 0.71.0)

benbrandt:
## Summary

Since #1004 (shipped in 0.71.0), `session/request_permission` for `Bash`/`PowerShell` replaced the tool-call title with the model-authored `input.description`, so both `toolCall.title` and `_meta.permission.title` read "List files in current directory" while the command being approved was not rendered anywhere in the prompt. `toolInfoFromToolUse` uses the command as the Bash title, so the override created exactly the divergence the comment above it ruled out.

- Drop the override: the permission heading is the standard tool-call title again, i.e. the command for Bash. This restores the pre-0.71.0 behavior.
- For shell tools, `_meta.permission.title` mirrors the tool-call title verbatim. Single-line normalization would rewrite quoted arguments and join lines, and the 4,000-character cap would drop the heading for the long commands that most need reading.
- The description is still available to clients as `_meta.claudeCode.title` on the standard `tool_call`, and as the card body when the client has no terminal support.
- Update `docs/permission-extension.md`; rewrite the tests that locked in the description title; add no-command fallback and whitespace/length fidelity coverage.

Fixes #1068

Related: #1070 makes the same core change but also routes PowerShell through the Bash terminal lifecycle in `tools.ts`. That expansion is left out here to keep the fix minimal; PowerShell prompts keep the pre-0.71.0 tool-name title.

## Tests

- `npm run check`
- `npm run build`
- `npm run test:run` (1371 passed, 28 skipped)


benbrandt:
Superseded by #1070, which makes the same presentation change and adds PowerShell parity. Rebased that PR onto main instead.

